PR #2, opened 2018-08-01
Add PermissionStore with tests

sisou:


src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.instance = this.instance || new PermissionStore();
+        return this.instance;
+    }
+
+    private constructor() {
+        this.dbPromise = null;
+        this.indexedDB = PermissionStore.INDEXEDDB_IMPLEMENTATION;
+    }
+
+    public allowByOrigin(origin: string, addresses: string|string[]): Promise<string> {
```
paberr: I think we should leverage the advantages of a typed language and use `Nimiq.Address` here.

src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
...
+                : Array.isArray(addresses)
+                    ? addresses
+                    : [addresses],
+        };
+
+        return this._put(permission);
+    }
+
+    public getByOrigin(origin: string): Promise<Permission> {
+        return this._get(origin);
```
paberr: Why do you need the private methods?

src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    public removeByOrigin(origin: string): Promise<true> {
+        return this._remove(origin);
```
paberr: Why do you need the private methods?

src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    public list(): Promise<Permission[]> {
+        return this._list();
```
paberr: Why do you need the private methods?

src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
...
+
+                        // Store values in the newly created objectStore.
+                        const permissionObjectStore = db.transaction([PermissionStore.DB_STORE_NAME], 'readwrite')
+                            .objectStore(PermissionStore.DB_STORE_NAME);
+
+                        defaultPermissions.forEach((permission) => permissionObjectStore.add(permission));
+                    };
+                }
+
+                // if (event.oldVersion < 2) {
```
paberr: I would remove this comment.

src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
...
+
+                if (event.oldVersion < 1) {
+                    // Version 1 is the first version of the database.
+                    const store = db.createObjectStore(PermissionStore.DB_STORE_NAME, { keyPath: 'origin' });
+
+                    // Add default permissions
+                    store.transaction.oncomplete = () => {
+                        // FIXME Make these origins dependent on the current environment
+                        // (localhost, testing, production)
+                        const nimiqOrigins = [
```
paberr: I would move this constant somewhere else and not hide it inside the code.

src/lib/PermissionStore.ts
paberr: I just don't know how to best handle returning `Permission` later (which should have `Nimiq.Address` in there as well).

src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
+export interface Permission {
+    origin: string;
+    allowsAll: boolean;
+    addresses: string[];
```
paberr: I think we should leverage the advantages of a typed language and use `Nimiq.Address` here.

src/lib/PermissionStore.ts
```diff
@@ -0,0 +1,165 @@
...
+                        // FIXME Make these origins dependent on the current environment
+                        // (localhost, testing, production)
+                        const nimiqOrigins = [
+                            'https://safe.nimiq-testnet.com',
+                            'https://miner.nimiq-testnet.com',
+                            'https://promo.nimiq-testnet.com',
+                            'https://shop.nimiq-testnet.com',
+                        ];
+
+                        const defaultPermissions = nimiqOrigins.map((origin) => ({
```
paberr: Maybe extracting the permission into a constant somewhere?

paberr:
Ah, another question: The linter complained about private variables starting with `_`. What should be our convention there?

Bettelstab:
I disabled that rule in my branch. IMHO it's nice to have the _ anyway so the information is not lost when compiling.

src/lib/Config.ts
```diff
@@ -0,0 +1,120 @@
...
+        return Config._origin(subdomain);
+    }
+
+    public static src(subdomain: string) {
+        return Config._origin(subdomain, true);
+    }
+
+    public static get tld() {
+        const tld = Config._myOrigin.split('.');
+        return [tld[tld.length - 2], tld[tld.length - 1]].join('.');
```
paberr: Technically, the TLD is only the last part of the origin, e.g., `.com` or `.de`.
But besides the naming, there is another problem when it comes to TLDs like `.co.uk`.
`test.co.uk` and `evil.co.uk` would yield the same result here!

paberr: As far as I can see, this will only be invoked from our domains (currently under the `.com` TLD).
So, I would at least add a comment describing the problem here.

src/lib/Config.ts
```diff
@@ -0,0 +1,120 @@
...
+    private static _origin(subdomain: string, withPath?: boolean) {
+        if (Config._myOrigin.includes('localhost')) {
+            return Config._localhost(subdomain, withPath);
+        }
+
+        if (Config.devMode) {
+            return Config._localhost(subdomain, withPath, true);
+        }
+
+        return `https://${subdomain}.${Config.tld}${(withPath && '/') || ''}`;
```
paberr: I would use the ternary if there: `withPath ? '/' : ''`. I think it is much easier to understand.

src/lib/Config.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        switch (Config.tld) {
+            case 'nimiq.com': return 'main';
+            case 'nimiq-testnet.com': return 'test';
+            default: return 'test'; // Set this to 'test', 'bounty', or 'dev' for localhost development
+        }
+    }
+
+    public static get cdn() {
+        if (Config.offlinePackaged) { return Config.src('keyguard') + '/nimiq.js'; }
```
paberr: I don't understand the rationale for `keyguard` here?
Perhaps it is just because of the current offline structure?

src/lib/Config.ts
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        return `https://${subdomain}.${Config.tld}${(withPath && '/') || ''}`;
+    }
+
+    private static _localhost(subdomain: string, withPath?: boolean, ipMode?: boolean) {
+        let path = '';
+
+        if (withPath) {
+            if (Config.offlinePackaged) { path = '/' + subdomain + '/'; } else {
```
paberr: I suggest putting newlines here to improve readability of the if-block.

src/lib/Config.ts
```diff
@@ -0,0 +1,120 @@
...
+                switch (subdomain) {
+                    case 'keyguard': path = '/libraries/keyguard/'; break;
+                    case 'network': path = '/libraries/network/'; break;
+                }
+
+                if (location.pathname.includes('/dist')) {
```
paberr: I suppose this is also because of the current localhost project structure, right?

src/lib/Config.ts
sisou: Yes, this was from how the offline package was packaged for the current Safe, where the Nimiq lib was included in the Keyguard folder.

Issue #8, opened 2018-08-15
Identicons don't load


Bettelstab:
(vue-components)

sisou:
Fixed in 7985dba05f0e52ee79f5aeeaa079abf95aeac26b (on a feature branch, not yet merged to master).

sisou:
Fixed by merging #11.

PR #11, opened 2018-08-16
Fix #4 - Checkout overview and success screens

sisou:
Fixes #4, #8.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -2,9 +2,10 @@ import Vue from 'vue';
 import Vuex, { StoreOptions } from 'vuex';
 import {ParsedRpcRequest} from '@/lib/RequestTypes';
 import {KeyInfo} from '@/lib/KeyInfo';
-import {State as RpcState} from '@nimiq/rpc';
+import {State as RpcState, State} from '@nimiq/rpc';
 import {KeyStore} from '@/lib/KeyStore';
 import {RpcResult as KeyguardResult} from '@/lib/keyguard/RequestTypes';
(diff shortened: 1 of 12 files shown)


src/store.ts
```diff
@@ -2,48 +2,107 @@ import Vue from 'vue';
...
+    activeAccountPath: string | null;
 }
 
 const store: StoreOptions<RootState> = {
-    state: {
-        rpcState: null, // undefined is not reactive
-        keys: [],
-        keyguardResult: null, // undefined is not reactive
-    },
+    state: (() => {
```
paberr: Is that the recommended way of doing that?
I would have rather put that into the `created` method similar to the initKeys.

src/views/CheckoutSelectAccount.vue
```diff
@@ -0,0 +1,128 @@
...
+            accountPath: addressInfo.path,
+        });
+
+        // Return to overview
+        this.backToOverview();
+    }
+
+    private backToOverview() {
+        this.$router.push({name: RequestType.CHECKOUT});
+    }
```
Bettelstab: I'd say anything but the backToOverview() is generic for all flows with login/account selector, right?

src/store.ts
```diff
@@ -2,48 +2,107 @@ import Vue from 'vue';
...
+        // Try loading active
+        let activeLoginId: null|string = null;
```
paberr: I generally prefer to format it like `string | null` (spaces and sorted by relevance).

PR #12, opened 2018-08-17
Notify client when window is closed

Bettelstab:


src/store.ts
```diff
@@ -43,6 +48,13 @@ const store: StoreOptions<RootState> = {
                 commit('initKeys', keys);
             });
         },
+        setKeyguardResult({ commit }, result) {
+            commit('setKeyguardResult', result);
+            commit('setWaitingForKeyguard', false);
```
sisou: Is this necessary? I think the state variable is always initialized as `false` when the page is loaded (which is the case, when returning from the keyguard).

Also, this is just two commits without any other actions performed, so could also be a mutation directly, I'd say?

sisou:
To enable the keyguard to throw the closing exception as well, the account manager needs to forward it the RPC request message ID, so that the caller can match the exception to the request.

src/views/Checkout.vue
```diff
@@ -33,7 +33,7 @@ import {AccountSelector, LoginSelector, PaymentInfoLine, SmallPage} from '@nimiq
 import {AddressInfo} from '../lib/AddressInfo';
 import {KeyInfo, KeyStorageType} from '../lib/KeyInfo';
 import {ParsedCheckoutRequest} from '../lib/RequestTypes';
-import {State, Mutation, Getter} from 'vuex-class';
+import {State, Mutation, Action, Getter} from 'vuex-class';
```
sisou: The Action import is now unnecessary ;)

PR #26, opened 2018-09-20
Refactor SignTransaction and Checkout success screens

sisou:
To enable correct routing when returning from Keyguard and to move transaction creation and sending logic into a shared `TransactionSender` component. This component currently only assembles the transaction and replies it to the RPC client, but will in the future be responsible for relaying the transaction to a server or creating the network consensus to send directly.

### DIFF
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -96,7 +96,7 @@ export default class RpcApi {
                 result.kind = command;
                 this._store.commit('setKeyguardResult', result);
 
-                this._router.push({name: keyguardResponseRouter[command].resolve});
+                this._router.push({name: keyguardResponseRouter(command, this._staticStore.request!.kind).resolve});
             }, (error, state) => {
                 // Recover state
                 this._recoverState(state);
@@ -104,7 +104,7 @@ export default class RpcApi {
                 // Set result
                 this._store.commit('setKeyguardResult', error);
 
-                this._router.push({name: keyguardResponseRouter[command].reject});
+                this._router.push({name: keyguardResponseRouter(command, this._staticStore.request!.kind).reject});
             });
         }
     }
--- a/src/views/Checkout.vue
+++ b/src/views/Checkout.vue
@@ -13,107 +13,16 @@
 </template>
 
 <script lang="ts">
-import {Component, Emit, Prop, Watch, Vue} from 'vue-property-decorator';
-import {AccountSelector, LoginSelector, PaymentInfoLine, SmallPage} from '@nimiq/vue-components';
-import {
-    RequestType,
-    ParsedCheckoutRequest,
-    ParsedSignTransactionRequest,
-    SignTransactionResult,
-} from '../lib/RequestTypes';
-import {AddressInfo} from '../lib/AddressInfo';
-import {KeyInfo, KeyStorageType} from '../lib/KeyInfo';
-import {State, Mutation, Getter} from 'vuex-class';
-import RpcApi from '../lib/RpcApi';
-import {
-    SignTransactionRequest as KSignTransactionRequest,
-    SignTransactionResult as KSignTransactionResult,
-} from '@nimiq/keyguard-client';
-import {ResponseStatus, State as RpcState} from '@nimiq/rpc';
+import {Component, Emit, Vue} from 'vue-property-decorator';
+import {PaymentInfoLine, SmallPage} from '@nimiq/vue-components';
+import {ParsedCheckoutRequest} from '../lib/RequestTypes';
+import {State as RpcState} from '@nimiq/rpc';
 import {Static} from '../lib/StaticStore';
 
 @Component({components: {PaymentInfoLine, SmallPage}})
 export default class Checkout extends Vue {
-    @State private keyguardResult!: KSignTransactionResult | Error | null;
-
     @Static private rpcState!: RpcState;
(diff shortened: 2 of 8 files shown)


src/lib/RpcApi.ts
```diff
@@ -96,15 +96,15 @@ export default class RpcApi {
                 result.kind = command;
                 this._store.commit('setKeyguardResult', result);
 
-                this._router.push({name: keyguardResponseRouter[command].resolve});
+                this._router.push({name: keyguardResponseRouter(command, this._staticStore.request!.kind).resolve});
```
sisou: To enable the `keyguardResponseRouter` to decide correctly to which route it should direct when returning from the Keyguard's sign-transaction request, the original request `kind` that was given to the AccountsManager is passed here and the `keyguardResponseRouter` is turned from an object into a function instead.

src/views/Checkout.vue
```diff
@@ -13,107 +13,16 @@
...
-    private async onKeyguardResult() {
-        if (this.keyguardResult instanceof Error) {
-            //
-        } else if (this.keyguardResult) {
-            // Load web assembly encryption library into browser (if supported)
-            await Nimiq.WasmHelper.doImportBrowser();
-            // Configure to use test net for now
-            Nimiq.GenesisConfig.test();
-
-            let tx: Nimiq.Transaction;
```
sisou: The whole transaction-creation-code has been moved to the central `TransactionSender` component.

src/views/CheckoutOverview.vue
```diff
@@ -18,9 +18,9 @@
 
 <script lang="ts">
```
sisou: This file just contains name refactorings, no logic changes.

src/views/CheckoutSelectAccount.vue
```diff
@@ -26,7 +26,7 @@ import {RequestType} from '../lib/RequestTypes';
 import {State, Mutation} from 'vuex-class';
 
 @Component({components: {AccountSelector, LoginSelector}})
-export default class Checkout extends Vue {
+export default class CheckoutSelectAccount extends Vue {
```
sisou: Only a name refactor with no external dependencies, because it's `export default`.

src/views/CheckoutSuccess.vue
```diff
@@ -1,27 +1,30 @@
 <template>
     <div class="success center">
         <div class="icon-checkmark-circle"></div>
-        <h1>Your {{ requestKindString }}<br>was successfull!</h1>
+        <h1>Your payment<br>was successfull!</h1>
```
sisou: Since both types of transactions now have a separate success component, the dynamic type is here not needed anymore.

src/views/CheckoutSuccess.vue
```diff
@@ -1,27 +1,30 @@
...
         <div style="flex-grow: 1;"></div>
-        <button @click="close">Back to store</button>
+        <button @click="close" :disabled="!isTxSent">Back to store</button>
```
sisou: Disable button until `TransactionSender` fires the `transaction-sent` event. This is only 1-2 seconds currently as it's loading the Nimiq library, but will be longer when this component needs to relay the tx to an external server or establish consensus in the future.

src/views/SignTransaction.vue
```diff
@@ -29,13 +29,16 @@ export default class SignTransaction extends Vue {
     @State private keyguardResult!: KSignTransactionResult | Error | null;
 
     public async created() {
-        if (this.$route.name === `sign-transaction-success`) return;
+        if (this.keyguardResult instanceof Error) {
+            this.rpcState.reply(ResponseStatus.ERROR, this.keyguardResult);
```
sisou: Since the sign-transaction flow does not currently have a landing page in the AccountsManager, there is also no place to display an error or have the user try again. Thus we forward the error directly to the caller (the Safe, for example) which automatically closes the window from the caller side.

src/views/SignTransaction.vue
```diff
@@ -29,13 +29,16 @@ export default class SignTransaction extends Vue {
...
+        } else if (this.keyguardResult) return; // Keyguard success is handled in SignTransactionSuccess.vue
```
sisou: The else condition here covers all successful signature cases. If the keyguardResult is successful, we do not want to restart the Keyguard request. The sending of the tx is handled in the `SignTransactionSuccess` component.

src/views/SignTransactionSuccess.vue
```diff
@@ -0,0 +1,71 @@
+<template>
```
sisou: Basically a copy of `CheckoutSuccess.vue` with only minor wording and button label differences. In the future, `SignTransaction` and `Checkout` flows are potentially handled more different though (sending tx to central server or directly into network during checkout, as opposed to returning it to the Safe for a regular signature request), thus it makes sense to separate the two.

Bettelstab:
Suggestion: Use some of Sören's remarks as code comments.

src/lib/RpcApi.ts
Bettelstab: This could be a good comment.

src/views/SignTransaction.vue
Bettelstab: Could also be a nice preliminary comment.

src/views/SignTransaction.vue
Bettelstab: Another useful comment ;)

PR #34, opened 2018-10-09
Sebastian/remove key

### LINKED ISSUE #25: Logout flow
Enable the user to "log out", to remove a wallet from their device.

nibhar:
Resolves #25 

needs github/nimiq/keyguard-next#100

### DIFF
--- a/src/lib/RequestTypes.ts
+++ b/src/lib/RequestTypes.ts
@@ -4,6 +4,7 @@ export enum RequestType {
     SIGNTRANSACTION = 'sign-transaction',
     SIGNUP = 'signup',
     LOGIN = 'login',
+    LOGOUT = 'logout',
 }
 
 export interface SignTransactionRequest {
@@ -117,15 +118,31 @@ export interface LoginResult {
     }>;
 }
 
+export interface LogoutRequest {
+    kind?: RequestType.LOGOUT;
+    appName: string;
+    keyId: string;
+    keyLabel?: string;
+}
+
+export interface ParsedLogoutRequest {
+    kind: RequestType.LOGOUT;
+    appName: string;
+    keyId: string;
+    keyLabel?: string;
+}
+
 // Discriminated Unions
 export type RpcRequest = SignTransactionRequest
                        | CheckoutRequest
                        | SignupRequest
-                       | LoginRequest;
+                       | LoginRequest
+                       | LogoutRequest;
 export type ParsedRpcRequest = ParsedSignTransactionRequest
                              | ParsedCheckoutRequest
                              | ParsedSignupRequest
-                             | ParsedLoginRequest;
+                             | ParsedLoginRequest
+                             | ParsedLogoutRequest;
 
 export class AccountsRequest {
     public static parse(request: RpcRequest, requestType?: RequestType): ParsedRpcRequest | null {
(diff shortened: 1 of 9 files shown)


src/lib/RequestTypes.ts
```diff
@@ -221,6 +250,13 @@ export class AccountsRequest {
                     kind: RequestType.LOGIN,
                     appName: request.appName,
                 } as LoginRequest;
+            case RequestType.LOGOUT:
+            return {
+                kind: RequestType.LOGOUT,
+                appName: request.appName,
+                keyId: request.keyId,
+                keyLabel: request.keyLabel,
+            } as LogoutRequest;
```
sisou: I guess you can just `return request` here instead of re-constructing it, because there no change from parsed to raw request.

src/lib/RequestTypes.ts
```diff
@@ -117,15 +118,35 @@ export interface LoginResult {
     }>;
 }
 
+export interface LogoutRequest {
+    kind?: RequestType.LOGOUT;
+    appName: string;
+    keyId: string;
+    keyLabel?: string;
```
sisou: `keyLabel` should not be accepted from outside, but just be filled in by the AccountsManager.

src/views/Logout.vue
```diff
@@ -0,0 +1,95 @@
...
+    @Static private request!: ParsedLogoutRequest;
+    @State private keyguardResult!: RemoveKeyResult | Error | null;
+
+    public async created() {
+        if (this.keyguardResult instanceof Error) {
+            this.rpcState.reply(ResponseStatus.ERROR, this.keyguardResult);
+        } else if (this.keyguardResult) return; // Keyguard success is handled in LogoutSuccess.vue
+
+        var label = this.request.keyLabel;
+        if (!this.request.keyLabel) {
```
sisou: `keyLabel` should not be accepted from outside, but just be filled in by the AccountsManager.

src/views/LogoutSuccess.vue
```diff
@@ -0,0 +1,75 @@
...
+export default class LogoutSuccess extends Vue {
+    @Static private request!: ParsedLogoutRequest;
+    @Static private rpcState!: RpcState;
+    @State private keyguardResult!: RemoveKeyResult;
+
+    public mounted() {
+        if (this.keyguardResult.success === true) {
+            KeyStore.Instance.remove(this.request.keyId);
+        }
+        this.rpcState.reply(ResponseStatus.OK, this.keyguardResult);
```
sisou: The page should only return to the caller when the user clicks the button. The deletion of the database entry should happen immediately, but the `reply` only when the `Back to <appName>` button is clicked. Thus the reply should be moved into the `close()` method to replace the `window.close()`.

src/views/LogoutSuccess.vue
```diff
@@ -28,11 +28,11 @@ export default class LogoutSuccess extends Vue {
         if (this.keyguardResult.success === true) {
             KeyStore.Instance.remove(this.request.keyId);
         }
-        this.rpcState.reply(ResponseStatus.OK, this.keyguardResult);
     }
 
     @Emit()
     private close() {
+        this.rpcState.reply(ResponseStatus.OK, this.keyguardResult);
         window.close();
```
sisou: `window.close()` needs to be removed, because it won't work for redirect requests. The popup is automatically closed by the caller when a response is received.

sisou: `window.close()` needs to be removed, because it won't work for redirect requests. The popup is automatically closed by the caller when a response is received.

PR #37, opened 2018-10-15
Detect and display accounts after key import

sisou:
Refactors the former `TransactionSender.vue` component (now `Network.vue`) to be dump and not use the `Static` or `State` stores. This requires logic to be put back into `SignTransactionSuccess` and `CheckoutSuccess`, the later of which now also automatically relays the transaction to the network before also returning it to the caller.

To facilitate the interaction from the AccountManager to the Keyguard iframe, I had to also adapt the `KeyguardClient` a fair bit, because it had no support for PostMessage behavior before `v.0.0.17`.

As a byproduct of testing, I also added a configurable `validityStartHeight` to the `sign-transaction` demo and hard coded the current frozen testnet height into the `checkout` demo.


src/components/Network.vue
```diff
@@ -0,0 +1,216 @@
...
+
+    private created() {
+        this.$on('consensus-established', () => this.consensusEstablished = true);
+        this.$on('consensus-lost', () => this.consensusEstablished = false);
+    }
+
+    private async _getNetworkClient(): Promise<NetworkClient> {
+        if (this._networkClient) return this._networkClient;
+
+        this._networkClient = new NetworkClient('http://localhost:5000');
```
sisou: This needs to be adapted to the respective development environment, or we define a standard where we always start our network locally. I think we cannot use the deployed network from our local environment, because the deployment uses HTTPS which we cannot communicate to and from from HTTP localhost.

src/views/CheckoutSuccess.vue
```diff
@@ -1,34 +1,44 @@
...
-    private isTxPrepared: boolean = false;
+    private result?: SignTransactionResult;
+    private isTxSent: boolean = false;
 
     private async mounted() {
-        await (this.$refs.txSender as TransactionSender).prepare();
-        // this.isTxPrepared = true;
-        (this.$refs.txSender as TransactionSender).send();
+        const tx = await (this.$refs.network as Network).prepareTx(this.keyguardRequest, this.keyguardResult);
+        this.result = await (this.$refs.network as Network).sendToNetwork(tx);
```
nibhar: Isn't this the place where the blocking until the event fires happens?

src/views/SignTransactionSuccess.vue
```diff
@@ -4,30 +4,40 @@
         <h1>Your transaction<br>is ready!</h1>
         <div style="flex-grow: 1;"></div>
         <button @click="done" :disabled="!isTxPrepared">Send now</button>
```
nibhar: This Button does not send the transaction but gives the SignTransactionResult to the caller. What he does with it is out of reach.

sisou: You are right, it needs to be relabelled. But then again, this whole page has to be changed to adhere to the designs. But you are still right.

src/views/CheckoutSuccess.vue
sisou: Yes. Is that a statement or a question?

src/views/CheckoutSuccess.vue
```diff
@@ -1,34 +1,44 @@
 <template>
     <div class="success center">
         <div class="icon-checkmark-circle"></div>
-        <h1>Your payment<br>was successfull!</h1>
+        <h1>Your payment<br>is beeing sent.<br>Please wait...</h1>
         <div style="flex-grow: 1;"></div>
-        <button @click="done" :disabled="!isTxPrepared">Back to store</button>
-        <TransactionSender ref="txSender"/>
+        <button @click="done" :disabled="!isTxSent">Back to store</button>
```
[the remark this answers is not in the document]
sisou: I agree. This page needs to be redone to be more like the designs anyway.

src/components/Network.vue
```diff
@@ -0,0 +1,216 @@
...
+     */
+    public async sendToNetwork(tx: Nimiq.Transaction): Promise<SignTransactionResult> {
+        const txObj = this.makeSignTransactionResult(tx);
+        const client = await this._getNetworkClient();
+
+        client.relayTransaction(txObj);
+
+        return new Promise<SignTransactionResult>((resolve, reject) => {
+            this.$once('transaction-relayed', (txInfo: any) => {
+                if (txInfo.hash === txObj.hash) resolve(txObj);
```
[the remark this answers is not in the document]
sisou: It is guaranteed in that if the transaction is really relayed, the event is guaranteed to fire. But you are right, this needs a timeout.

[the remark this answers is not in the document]
nibhar: so as you said in the other conversation, this does in fact not block until the event fires, which the comment above it suggests. In case that it really only should return once (if) the event fires this seems not sufficient.

PR #46, opened 2018-11-05
use props instead of vue instance for simple requests

nibhar:
Have a single Success Component instead of multiples (eg ExportWordsSuccess, ExportFileSuccess, ...). Use props for different wording.


sisou:
If the success component is used as a component and not as a view, it should be moved to `src/components`.

sisou:
But as a component, it cannot have access to the `keyguardResult` or request or to the `rpcState`.

Bettelstab:
> But as a component, it cannot have access to the `keyguardResult` or request or to the `rpcState`.

Why not?

sisou:
> Why not?

Because components should be dumb and not have dependencies on global state. They can only get outside info via props.

Bettelstab:
Reference? IMO that's a (not so uncommon ) misconception. The conceptual difference between view and component is that a view is routable.

Imagine you have an app which a header bar which shows content based on your login state (profile name, new messages... or login form). You can realise it as a component contained in every view (possibly having a base layout class), but you wouldn't pass all the store information from the view to this component as props. Instead... you just grab it from global state in your header bar component. ;)

PR #50, opened 2018-11-06
Simplify storing state during redirect requests

sisou:


client/AccountsManagerClient.ts
```diff
@@ -42,12 +43,20 @@ export default class AccountsManagerClient {
...
-        this._redirectClient.onResponse(command, resolve, reject);
+    public on(
+        command: RequestType,
+        resolve: (result: RpcResult, state: any) => any,
+        reject?: (error: Error, state: any) => any
+    ) {
+        this._redirectClient.onResponse(command,
+            // State is always an object containing at least the __command property
+            (result: RpcResult, rpcId, state) => resolve(result, JSON.parse(state!)),
+            (error: Error, rpcId, state) => reject && reject(error, JSON.parse(state!)),
```
svub: I think we agreed that state will always be a string now. I'm OK with changing though. Final?

demos/Demo.ts
```diff
@@ -145,7 +145,7 @@ class Demo {
         }
 
         function checkoutRedirect(txRequest: CheckoutRequest) {
-            return client.checkout(txRequest, new RedirectRequestBehavior());
+            return client.checkout(txRequest, new RedirectRequestBehavior(null, {storedGreeting: 'Hello Nimiq!'}));
```
svub:
```suggestion
            return client.checkout(txRequest, new RedirectRequestBehavior(null, { storedGreeting: 'Hello Nimiq!' }));
```
:)

svub: to further simplify the API, one could also check if the first parameter is an object and thus assume = (null, state); But I think we hardly do any of this "parameter magic", right?

sisou: Yeah, I don' think we do that anywhere. But if it becomes a common pain point, we can introduce it in the future.

client/AccountsManagerClient.ts
sisou: I found out that it actually cannot be a string, specifically because of this line:
https://github.com/nimiq/accounts/blob/master/client/RequestBehavior.ts#L58

I talked with Pascal and he agreed that it won't work, thus I decided to just go with a plain object instead and added the de-serialization wrapper.

PR #49, opened 2018-11-06
Refactor names

sisou:
Key => Wallet, Address => Account, cleanup a few unused imports

PR #52, opened 2018-11-09
Documentation and naming refactoring

sisou:
This is the docs changes, but also the changes from #49 (including the whitespace linting).


sisou:
Please be advised that this PR also renames the Accounts IndexedDB and thus all stored wallets are lost and need to be re-imported. There is no use in handling migration now, since it's only the testnet. If you are worried about losing your testnet coins, please backup recovery words of all wallets while still on the `master` branch before this PR gets merged.

PR #57, opened 2018-11-17
Add empty templates to view components

sisou:
Vue is giving me this warning when handing through to the Keyguard:

`[Vue warn]: Failed to mount component: template or render function not defined.`

I think adding empty templates doesn't hurt, what do you think?


svub:
Should be ok if it doesn't come with any overhead or performance penalty.

PR #61, opened 2018-11-25
Rename request

### LINKED ISSUE #24: [User flow] Rename Wallets and Accounts
Enable the user to rename wallets and accounts.

nibhar:
resolves #24 

Enables the user to rename his Wallets and Accounts. Does currently not require the passphrase.

The correct account/wallet that was clicked to be renamed originally will be pre selected once the page loaded. All other accounts of that wallet and the wallet itself can then also be renamed so that renaming multiple Accounts does only take 1 request per Wallet.

### DIFF
--- a/src/views/Rename.vue
+++ b/src/views/Rename.vue
@@ -1,6 +1,6 @@
 <template>
     <div class="container pad-bottom">
-        <SmallPage>
+        <SmallPage class="rename">
             <PageHeader>Rename your Wallet</PageHeader>
             <PageBody v-if="wallet">
                 <div class="wallet-label">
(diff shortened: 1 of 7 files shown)


src/views/Rename.vue
```diff
@@ -0,0 +1,210 @@
+<template>
+    <div class="container">
+        <SmallPage class="rename">
+            <PageHeader>Rename your Wallet</PageHeader>
+            <PageBody v-if="wallet">
+                <div class="wallet-label">
+                    <div class="wallet-icon nq-icon" :class="walletIconClass"></div>
+                    <LabelInput :value="wallet.label" @changed="onWalletLabelChange" ref="wallet"/>
+                </div>
```
sisou: For legacy wallets the wallet label is not relevant and should be hidden.

PR #65, opened 2018-11-26
Enable importing wallets from ActiveAccountSelector during checkout

### LINKED ISSUE #19: Guide user through login with new key during checkout
The first page of the checkout flow enables the user to select which login and account to use to pay. The case where no key is yet stored, is not yet covered.

A way to let the user import a key (from file or words) needs to be provided, after which the checkout flow should proceed regularily.

sisou:
Resolves #19.

### DIFF
--- a/src/views/ActiveAccountSelector.vue
+++ b/src/views/ActiveAccountSelector.vue
@@ -59,8 +59,6 @@ export default class ActiveAccountSelector extends Vue {
     }
 
     private switchWallet() {
-        // TODO Redirect to import/create just like addWallet()?
-
         this.page = 1;
     }
 
--- a/src/views/ActiveAccountSelector.vue
+++ b/src/views/ActiveAccountSelector.vue
@@ -24,13 +24,11 @@ import { Component, Emit, Vue } from 'vue-property-decorator';
 import { AccountSelector, WalletSelector } from '@nimiq/vue-components';
 import { AccountInfo } from '../lib/AccountInfo';
 import { WalletInfo, WalletType } from '../lib/WalletInfo';
-import { State, Mutation } from 'vuex-class';
+import { State } from 'vuex-class';
 
 @Component({components: {AccountSelector, WalletSelector}})
 export default class ActiveAccountSelector extends Vue {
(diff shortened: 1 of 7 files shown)


src/views/Checkout.vue
```diff
@@ -18,13 +18,58 @@
...
+
+    private addWallet() {
+        // Redirect to import
+        const request: KeyguardRequest.ImportRequest = {
+            appName: this.request.appName,
+            defaultKeyPath: `m/44'/242'/0'/0'`,
+            requestedKeyPaths: [`m/44'/242'/0'/0'`],
+        };
+
+        staticStore.originalRequestType = RequestType.CHECKOUT;
```
nibhar: As this is used exclusively as a route later I suggest renaming it that and store the actual route name 
```suggestion
        staticStore.originalRouteName = RequestType.CHECKOUT;
```
There is potential for the originalRoute not being RequestName, but maybe further into the request.

src/views/LoginSuccess.vue
```diff
@@ -229,6 +229,14 @@ export default class LoginSuccess extends Vue {
         if (this.accountsUpdateCount) return Array.from(this.accounts.values());
         return [];
     }
+
+    private get appName() {
+        console.log(staticStore);
```
nibhar: can be deleted.

src/views/LoginSuccess.vue
```diff
@@ -229,6 +229,14 @@ export default class LoginSuccess extends Vue {
...
+        if (staticStore.originalRequestType) {
+            return staticStore.originalRequestType;
```
nibhar: This does not seem right.
The request types are not necessarily good language for display, neither are the routes. Also might cause some trouble when translating.

src/views/ActiveAccountSelector.vue
```diff
@@ -61,24 +59,14 @@ export default class ActiveAccountSelector extends Vue {
     }
 
     private switchWallet() {
-        // TODO Redirect to import/create just like addWallet()
+        // TODO Redirect to import/create just like addWallet()?
```
nibhar: I don't think there should be a redirect here.

PR #68, opened 2018-11-26
Add Sentry error reporting tool

sisou:
https://sentry.io

To add more team members, we need to use a paid plan, but it has a free trial. Just tell me if I should add you.

### DIFF
--- a/src/views/ErrorHandler.vue
+++ b/src/views/ErrorHandler.vue
@@ -3,6 +3,7 @@
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator';
 import { State } from 'vuex-class';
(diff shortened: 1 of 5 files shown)


src/views/ErrorHandler.vue
```diff
@@ -11,6 +12,8 @@ export default class ErrorHandler extends Vue {
     public async created() {
         if (this.keyguardResult instanceof Error) {
             if (!this.requestSpecificErrors()) {
+                Sentry.captureException(this.keyguardResult);
```
Bettelstab: Could it possibly make more sense to capture errors in $rpc.reject()?

PR #82, opened 2018-12-04
Enable triggering Keyguard migration

sisou:
For simplicity and reduction of potential errors, I decided to make the migration a dedicated step. So the AccountsClient will return a `MIGRATION_REQUIRED` error when detecting that no wallets exist, only legacy accounts.

The app (Safe) then has to trigger the `migrate` request, which opens an Accounts Manager window which does the migration. It does the Keyguard migration in an iframe and then closes automatically.

In the future, the `migrate` request can also be triggered internally in the AccountManager from any other request.

It will use the new Loader component that is being prepared in another branch.

For testing, https://github.com/nimiq/keyguard-next/pull/129 is required.


package.json
```diff
@@ -7,10 +7,11 @@
         "build": "vue-cli-service build --modern App",
         "lint": "vue-cli-service lint",
         "test:unit": "vue-cli-service test:unit",
-        "test": "yarn test:unit"
+        "test": "yarn test:unit",
+        "pr": "yarn lint && yarn build && yarn test && cd client && yarn lint && yarn build"
```
danimoh: built accounts client should be added to gitignore

src/views/Migrate.vue
```diff
@@ -0,0 +1,65 @@
...
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator';
+import { AccountInfo } from '@/lib/AccountInfo';
+import { WalletStore } from '@/lib/WalletStore';
+import { WalletInfo, WalletType } from '@/lib/WalletInfo';
+
+@Component
+export default class Migrate extends Vue {
+    private status: string = 'Connecting to Keyguard...';
+    private waitUntilReturn: number = 1000;
```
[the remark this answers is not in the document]
nibhar: we probably will need this very constant in some other places too. A couple of Success screens come to mind. So maybe we can think about a place where it makes sense for all of them to get it.

package.json
sisou: Depends. If we want to link to the build files for users to download the client, then it must be included. Otherwise the download has to be from somewhere else.

sisou: Tracked here: https://github.com/nimiq/accounts/pull/90.

PR #83, opened 2018-12-04
Add Loader component

sisou:
The new icons need to be extracted into @nimiq/style in the future when the icon set is ready.

### DIFF
--- a/src/components/Loader.vue
+++ b/src/components/Loader.vue
@@ -1,6 +1,6 @@
 <template>
     <div class="loader" :class="showLoadingBackground && (lightBlue ? 'nq-bg-light-blue' : 'nq-bg-blue')">
-        <transition name="fade-out">
+        <transition name="fade-loading">
             <div class="wrapper" v-if="state === 'loading'">
                 <h1 class="title nq-h1">{{ loadingTitle }}</h1>
 
@@ -15,7 +15,7 @@
             </div>
         </transition>
 
-        <transition name="fade-in">
+        <transition name="fade-result">
             <div class="wrapper success nq-bg-green" v-if="state === 'success'">
                 <div class="icon-row">
                     <div class="success nq-icon"></div>
@@ -26,7 +26,7 @@
             </div>
         </transition>
 
-        <transition name="fade-in">
+        <transition name="fade-result">
             <div class="wrapper warning nq-bg-orange" v-if="state === 'warning'">
                 <div class="top-spacer" :class="{'with-main-action': !!mainAction, 'with-alternative-action': !!alternativeAction}"></div>
 
@@ -44,7 +44,7 @@
             </div>
         </transition>
 
-        <transition name="fade-in">
+        <transition name="fade-result">
             <div class="wrapper error nq-bg-red" v-if="state === 'error'">
                 <div class="top-spacer" :class="{'with-main-action': !!mainAction, 'with-alternative-action': !!alternativeAction}"></div>
 
@@ -150,6 +150,10 @@ export default class Loader extends Vue {
         // When the state changes later and animates
         if (oldState && (newState !== Loader.Status.LOADING)) {
             setTimeout(() => this.showLoadingBackground = false, 1000);
+        }
+
+        if (newState === Loader.Status.LOADING) {
+            this.showLoadingBackground = true;
         }
     }
 
@@ -315,19 +319,20 @@ export default class Loader extends Vue {
 
     /* FADE transitions */
 
-    .fade-out-leave-active {
+    .fade-loading-leave-active,
+    .fade-result-leave-active {
         transition: opacity 300ms;
     }
 
-    .fade-out-leave-to {
-        opacity: 0;
-    }
-
-    .fade-in-enter-active {
+    .fade-loading-enter-active,
+    .fade-result-enter-active {
         transition: opacity 700ms 300ms;
     }
 
-    .fade-in-enter {
+    .fade-loading-enter,
+    .fade-loading-leave-to,
+    .fade-result-enter,
+    .fade-result-leave-to {
         opacity: 0;
     }
 </style>
--- a/src/components/Loader.vue
+++ b/src/components/Loader.vue
@@ -92,17 +92,27 @@ import { Component, Prop, Watch, Emit, Vue } from 'vue-property-decorator';
  *
  * **@alternative-action**
  *
- * The `state` is also available as Loader.LOADING, Loader.SUCCESS, Loader.WARNING and Loader.ERROR.
+ * The `state` is available as `Loader.Status.LOADING`, `Loader.Status.SUCCESS`,
+ * `Loader.Status.WARNING` and `Loader.Status.ERROR`.
+ *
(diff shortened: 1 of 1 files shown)


src/components/Loader.vue
```diff
@@ -0,0 +1,307 @@
+<!--
+Usage:
+
+<Loader
+    :title="isLoading ? 'Loading things' : 'Things done'"
+    :status="'[optional] Currently doing this'"
+    :state="[optional, default 'loading'] isLoading ? 'loading' : 'success'"
+    :lightBlue="[optional, default false] true"
+    :message="'[optional] But you can safely try again.'"
```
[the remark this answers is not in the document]
sisou: Do you think `explanation` would be better? Or `errorAndWarningText` (the last one is obviously a joke :smile:)?

src/components/Loader.vue
```diff
@@ -0,0 +1,307 @@
+<!--
+Usage:
+
+<Loader
+    :title="isLoading ? 'Loading things' : 'Things done'"
```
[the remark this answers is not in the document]
sisou: The external component is free to use whichever means they want to get a title in there. But I think in the use cases that we currently have, there is only the loading -> success case.

src/components/Loader.vue
```diff
@@ -0,0 +1,315 @@
...
+
+    @Watch('status', {immediate: true})
+    private updateStatus(newStatus: string, oldStatus: string) {
+        this.nextStatus = newStatus;
+        this.isStatusTransitioning = true;
+
+        setTimeout(() => {
+            this.currentStatus = newStatus;
+            this.isStatusTransitioning = false;
+        }, 500);
```
nibhar: what happens if i Change status before that timeout.

src/components/Loader.vue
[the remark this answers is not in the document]
Bettelstab: xD Well, I guess message with proper description like now is fine.

src/components/Loader.vue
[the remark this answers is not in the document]
Bettelstab: But in e.g. the direct error case, you also show the title. If the example is just for loading->success case, we don't need message at all. That's what I don't like about it. Either we should put a real example or just go with :title='the current title' and so on.

src/components/Loader.vue
sisou: I don't know, let's find out! :laughing:

src/components/Loader.vue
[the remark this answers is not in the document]
nibhar: yes. It will end up being a computed property most of the times. 
I don't think too much emphasis on that is required, as that is caller side. The component itself does not distinguish the title for the different state values.

src/components/Loader.vue
```diff
@@ -0,0 +1,323 @@
...
+ * **@alternative-action**
+ *
+ * The `state` is also available as Loader.LOADING, Loader.SUCCESS, Loader.WARNING and Loader.ERROR.
+ */
+@Component
+export default class Loader extends Vue {
+    private static LOADING = 'loading';
+    private static SUCCESS = 'success';
+    private static WARNING = 'warning';
+    private static ERROR   = 'error';
```
danimoh: An enum Loader.State would be nicer. Also should be `readonly`

src/components/Loader.vue
```diff
@@ -0,0 +1,323 @@
...
+    private updateState(newState: string, oldState: string) {
+        // When the component is initialized with a state other than LOADING
+        if (!oldState && (newState !== Loader.LOADING)) {
+            this.showLoadingBackground = false;
+        }
+
+        // When the state changes later and animates
+        if (oldState && (newState !== Loader.LOADING)) {
+            setTimeout(() => this.showLoadingBackground = false, 1000);
+        }
```
danimoh: If the state changes back to `LOADING`, `showLoadingBackground` needs to be reset to `true` and the timeout cancelled.

src/components/Loader.vue
```diff
@@ -0,0 +1,323 @@
...
+        this.isStatusTransitioning = true;
+
+        setTimeout(() => {
+            this.currentStatus = newStatus;
+            this.isStatusTransitioning = false;
+        }, 500);
+    }
+
+    private onMainAction() {
+        this.$emit('main-action');
```
danimoh: Events might be and enum also. I personally like to have it this way, because that way, one avoids typos and can see which events are available at one glance and also it enables useful IDE autocompletion.
Other opinions?

src/components/Loader.vue
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    private onAlternativeAction() {
+        this.$emit('alternative-action');
```
danimoh: same same

src/components/Loader.vue
sisou: The component is not meant to be switched back, but I see your point.

sisou: Resolved with 4e4d856.

src/components/Loader.vue
sisou: It works. Looks not optimal, but works for now and displays the correct status in the correct order.

src/components/Loader.vue
```diff
@@ -0,0 +1,315 @@
+<!--
+Usage:
+
+<Loader
+    :title="isLoading ? 'Loading things' : 'Things done'"
+    :status="'[optional] Currently doing this'"
+    :state="[optional, default 'loading'] isLoading ? 'loading' : 'success'"
+    :lightBlue="[optional, default false] true"
```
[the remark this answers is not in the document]
sisou: Explanation of the `lightBlue` prop has been added to the JSDoc.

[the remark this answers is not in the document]
nibhar: Is there a reason why we restrict ourselves to `nq-bg-blue` and `nq-bg-light-blue`?
So the default for lightBlue is to not show a light blue loading screen. What does it show then?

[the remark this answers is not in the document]
sisou: As per the designs, it uses the dark blue color by default. Green is reserved for success, orange for warning and red for error. You can get a green, orange or red screen without loading first as well, by initializing the component with the `state` you want.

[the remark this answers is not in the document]
sisou: image

src/components/Loader.vue
```diff
@@ -0,0 +1,342 @@
...
+
+namespace Loader { // tslint:disable-line no-namespace
+    export enum Status {
+        LOADING = 'loading',
+        SUCCESS = 'success',
+        WARNING = 'warning',
+        ERROR = 'error',
+    }
+
+    export enum Events {
```
nibhar: Singular or Plural.

danimoh: In the keyguard-next code we usually used `Events`

src/components/Loader.vue
danimoh: Not clearing the timeout though

src/components/Loader.vue
nibhar: Once the type is available it becomes somewhat strange as Singular for `enum`.
`const myCar: Cars = Cars.BlueCar;` does not read well.

PR #85, opened 2018-12-05
Handling of KEY_NOT_FOUND

### LINKED ISSUE #73: Read existing accounts in LoginSuccess
The login success screen currently overwrites an existing wallet and re-detects new accounts and assigns a default wallet label and account labels to an imported wallet. However, if a user reimports a wallet, his existing accounts and labels should be used instead of overwriting them.

### LINKED ISSUE #74: Handle non-existing Keyguard wallets
If a wallet was deleted in the Keyguard not through the accounts manager but for example by the browser, flag it as deleted in the accounts manager but keep the database entry such that the user doesn't lose his labels and can reimport the key with his old labels.

nibhar:
resolves #73 
resolves #74 

Upon encountering a KeyId that exist in Accounts but not in the Keyguard, the key gets flagged as deleted.
When Importing a key the existing labels will be preserved.


src/views/ErrorHandler.vue
```diff
@@ -3,19 +3,50 @@
...
 
     public async created() {
         if (!(this.keyguardResult instanceof Error)) return;
         if (this.requestSpecificErrors()) return;
-
-        // TODO proper Error Handling
-        // console.log(this.keyguardResult);
+        if (this.keyguardResult.message === Errors.Messages.KEY_NOT_FOUND) {
+            let walletId;
+            // ParsedExportRequest is just one Rrequest that has walletId. Any of those would do.
```
Bettelstab: fix typo :)

src/views/LoginSuccess.vue
```diff
@@ -165,7 +183,9 @@ export default class LoginSuccess extends Vue {
                 // It relieves us from having to calulate the correct array index with an
                 // unrelated counter variable.
                 pathsToDerive[this.lastDerivedIndex],
-                this.defaultAccountLabel,
+                (this.accounts.has(userFriendlyAddress))
+                    ? this.accounts.get(userFriendlyAddress)!.label
```
Bettelstab: Not sure, but there was a recent change to keys having this legacyAccount property now. Does it also apply here? If not, didn't we agree to use humanReadableAdress everywhere we can?

src/views/LoginSuccess.vue
```diff
@@ -66,21 +66,37 @@ export default class LoginSuccess extends Vue {
...
+                    addressObj.keyPath,
+                    this.defaultAccountLabel,
+                    address,
+                );
+                this.accounts.set(addressInfo.userFriendlyAddress, addressInfo);
+            }
         });
-
+        this.accountsUpdateCount += 1; // Trigger DOM update via computed property `this.accountsArray`
+        this.render = true; // trigger rendering now, when variables are set up correctly.
```
Bettelstab: I would prefer an adjective instead of 'render', like 'contentReady' or whatever.

src/views/ErrorHandler.vue
```diff
@@ -3,19 +3,50 @@
...
+                await WalletStore.Instance.put(walletInfo);
+            }
+            // TODO visuals
+            this.$rpc.reject(this.keyguardResult); // return it to caller
+            return;
+        }
+        // TODO more Error Handling
 
         this.$rpc.reject(this.keyguardResult);
+        return;
```
sisou: I think this `return` is unnecessary as it's at the end of the function anyway.

src/views/LoginSuccess.vue
```diff
@@ -5,15 +5,15 @@
                 <PageHeader>Your wallet is ready</PageHeader>
 
                 <PageBody>
-                    <div class="wallet-label" v-if="keyguardResult.keyType !== 0 /* LEGACY */">
+                    <div class="wallet-label" v-if="keyguardResult.keyType !== 0 /* LEGACY */ && render">
```
sisou: I'd put the check for `render` at the beginning of the `v-if` expression.

src/views/LoginSuccess.vue
```diff
@@ -66,21 +66,37 @@ export default class LoginSuccess extends Vue {
 
     private result?: LoginResult;
 
+    private render: boolean = false;
+    private retrievalComplete: boolean = false;
+
     private async mounted() {
+        const currentWalletInfo = await WalletStore.Instance.get(this.keyguardResult.keyId);
+
+        // if there is a WalletInfo for the id already, use it
```
```suggestion
        // if there is a WalletInfo for the key already, use it
```

src/views/LoginSuccess.vue
```diff
@@ -66,21 +66,37 @@ export default class LoginSuccess extends Vue {
...
+            this.walletLabel = currentWalletInfo.label;
+            if (currentWalletInfo.accounts) {
+                this.accounts = currentWalletInfo.accounts;
+            }
+        }
+
         // The Keyguard always returns (at least) one derived Address,
         // thus we can already create a complete WalletInfo object that
         // can be displayed while waiting for the network.
+        // before adding, make sure it is not set already
```
sisou: The existing comment uses proper capitalization and punctuation. Please do, too.

src/views/LoginSuccess.vue
sisou: The `legacyAccount` property is only relevant when migrating legacy account storage to new key storage. Thus not applicable here, as the Keyguard has no knowledge about account labels after the migration anymore.

Do you want to refactor `userFriendlyAddress => humanReadableAddress`? I don't remember the decision, but you may create an issue for it to track it ;).

tests/unit/CookieJar.spec.ts
```diff
@@ -41,6 +41,7 @@ const DUMMY_WALLET_OBJECTS: WalletInfoEntry[] = [
         ]),
         contracts: [],
         type: WalletType.BIP39,
+        deleted: false,
```
sisou: I think we should add encoding this flag to the CookieJar and test it here, but that can be a separate issue.

src/views/LoginSuccess.vue
Bettelstab: alright

src/lib/WalletInfo.ts
```diff
@@ -13,14 +13,15 @@ export class WalletInfo {
         o.accounts.forEach((accountInfoEntry, userFriendlyAddress) => {
             accounts.set(userFriendlyAddress, AccountInfo.fromObject(accountInfoEntry));
         });
-        return new WalletInfo(o.id, o.label, accounts, o.contracts, o.type);
+        return new WalletInfo(o.id, o.label, accounts, o.contracts, o.type, o.deleted);
```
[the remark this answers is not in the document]
nibhar: I went with keyMissing.

tests/unit/CookieJar.spec.ts
```diff
@@ -151,6 +151,7 @@ const BYTES = [
 
     // wallet 1 (BIP39)
     30, 227, 217, 38, 164, 156, // wallet id
+    0, // keyDeleted = false
```
danimoh: you apparently renamed this to `keyMissing`

tests/unit/CookieJar.spec.ts
nibhar: I put the first two `WalletInfoEntry` to `keyMissing=false,` and the last one to `keyMissing=true,` and Updated `BYTES` and `OUT_DUMMY_WALLET_OBJECTS` accordingly.

src/lib/CookieDecoder.ts
```diff
@@ -90,7 +94,7 @@ export class CookieDecoder {
             label: walletLabel,
             accounts,
             contracts: [],
-            deleted: false,
+            keyMissing: false,
```
sisou: You forgot to remove the default `false` here. But that this passes the test, means the test is broken. So first fix that the test reports an error here.

src/lib/CookieJar.ts
```diff
@@ -43,6 +44,11 @@ export default class CookieJar {
             const walletIdChunks = wallet.id.match(/.{2}/g);
             for (const chunk of walletIdChunks!) bytes.push(parseInt(chunk, 16));
 
+            // status
```
sisou: `status => Status`

src/lib/CookieDecoder.ts
```diff
@@ -47,6 +47,10 @@ export class CookieDecoder {
         let id: string = '';
         for (let i = 0; i < 6; i++) id += this.readByte(bytes).toString(16);
 
+        // Status byte
+        const statusByte = this.readByte(bytes);
+        const keyMissing = (statusByte & 1) === 1;
```
sisou: I'd say, to show that this is a binary setting, the `& 1` should be `& 0b1`.

src/lib/CookieJar.ts
```diff
@@ -43,6 +44,11 @@ export default class CookieJar {
...
+            let statusByte: number = 0;
+            statusByte = statusByte | (wallet.keyMissing ? 1 : 0);
```
sisou: I suggest using `0b1` and `0b0`.

src/lib/CookieDecoder.ts
nibhar: Bit flags are now available as enum (only `KEY_MISSING` for now)

src/lib/CookieJar.ts
nibhar: In this case it would be kind of alright to use `0b0`, but looking at the 7th bit it is just unnecessarily complex i think.

```suggestion
statusByte = statusByte | (wallet.keyMissing ? CookieJar.StatusFlags.KEY_MISSING : 0b0000000);
```
which in the end does exactly the same thing, but I think is even less understandable.

src/lib/CookieDecoder.ts
nibhar: The test works now. It tests 6 wallets now instead of 3 to cover every combination of KeyType and keyMissing Status.

src/lib/CookieJar.ts
```diff
@@ -136,3 +136,12 @@ export default class CookieJar {
         return Nimiq.BufferUtils.fromBase64(encodedWallets).length;
     }
 }
+
+namespace CookieJar { // tslint:disable-line no-namespace
+    export enum StatusFlags {
```
sisou: If we use the plural form, then the type is `CookieJar.StatusFlags`. I really think we should use more singular for these enums.

nibhar: In my opinion it is fine for bit flags, as you can have more than one ` CookieJar.StatusFlags.KEY_MISSING | CookieJar.StatusFlags.HAS_PIN`. Generally it makes sense to use singular though, especially if the type is used regularly (rarely for bit flags).

Issue #86, opened 2018-12-05
Proper error classes / constants


Bettelstab:

sisou:
Some constants currently in use:
```
LoginSuccess.DEFAULT_ACCOUNT_LABEL
LoginSuccess.KEYGUARD_BIP39_BASE_PATH
Loader.SUCCESS_REDIRECT_DELAY
```

nibhar:
`Somewhere.DEFAULT_WALLET_LABEL` // to be added

PR #91, opened 2018-12-07
move account finding to a library and refactor LoginSuccess

danimoh:
- create a library AccountFinder that orchestrates finding used accounts from derived accounts via the network
- refactor LoginSuccess to use AccountFinder and also clean up and structure the code and fix some smaller issues
- fix indentation in router.ts


src/router.ts
```diff
@@ -1,9 +1,11 @@
 import Vue from 'vue';
```
sisou: To enable proper code review, you HAVE TO remove the indendation changes from this PR or at least separate them out into a different commit, otherwise it's impossible to tell for us what changed and what hasn't.

danimoh: Just ignore whitespace: https://github.com/nimiq/accounts/pull/91/files?w=1

src/views/LoginSuccess.vue
```diff
@@ -36,9 +36,12 @@ import { WalletStore } from '@/lib/WalletStore';
 import staticStore, { Static } from '@/lib/StaticStore';
 import { PageHeader, PageBody, LabelInput, AccountList, PageFooter, SmallPage } from '@nimiq/vue-components';
 import Network from '@/components/Network.vue';
+import AccountFinder from '../lib/AccountFinder';
 
 @Component({components: {PageHeader, PageBody, LabelInput, AccountList, Network, PageFooter, SmallPage}})
 export default class LoginSuccess extends Vue {
+    private static readonly DEFAULT_ACCOUNT_LABEL: string = 'Standard Account';
```
nibhar: I"d say lets think about where to put those kind of constants. It gets used in other places too, so we should have a collection of them somehow.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+            const derivedAccounts = await derivedAccountsPromise;
+            if (derivedAccounts.length !== AccountFinder.MAX_ALLOWED_GAP) {
+                throw new Error('\'deriveAccounts\' returned wrong number of accounts');
+            }
+
+            // already start deriving next accounts
+            derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+            nextStartIndex += derivedAccounts.length; // by always advancing in blocks of MAX_ALLOWED_GAP it can happen
+            // that we include accounts that are more than MAX_ALLOWED_GAP apart (e.g. one towards the beginning of
+            // the first block, the other one towards the end of the second block). But that's actually not bad.
```
sisou: I don't understand this comment? Because in the comment you talk about `MAX_ALLOWED_GAP` while the code line actually adds `derivedAccounts.length`.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+    ): Promise<AccountInfo[]> {
+        await NetworkClient.Instance.init(); // initialize NetworkClient if not initialized yet
+        const foundAccounts: Array<{ address: string, keyPath: string, balance: number }> = [];
+        let nextStartIndex = 0;
+        let derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+        nextStartIndex += AccountFinder.MAX_ALLOWED_GAP;
+        while (true) {
+            const derivedAccounts = await derivedAccountsPromise;
+            if (derivedAccounts.length !== AccountFinder.MAX_ALLOWED_GAP) {
+                throw new Error('\'deriveAccounts\' returned wrong number of accounts');
```
sisou: Can we maybe handle this more gracefully instead of throwing, or is this an unexpected error that we need to know about if it happens?

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+            // already start deriving next accounts
+            derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+            nextStartIndex += derivedAccounts.length; // by always advancing in blocks of MAX_ALLOWED_GAP it can happen
+            // that we include accounts that are more than MAX_ALLOWED_GAP apart (e.g. one towards the beginning of
+            // the first block, the other one towards the end of the second block). But that's actually not bad.
+
+            // find accounts with a balance > 0
+            // TODO should use transaction receipts combined with balance
+            let didFindAccounts = false;
+            const userFriendlyAddresses = derivedAccounts.map(({ address }) => address);
```
sisou: Is object destructuring wideley enough supported? We are not using it anwhere else in the code base yet, so I suggest just using `account => account.address`.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
+import { AccountInfo } from '@/lib/AccountInfo';
+import { NetworkClient } from '@nimiq/network-client';
+import { WalletStore } from '@/lib/WalletStore';
+
+export default class AccountFinder {
+    public static async findAccounts(
+        deriveAccounts: (startIndex: number, count: number) => Promise<Array<{ address: string, keyPath: string }>>,
```
sisou: I suggest using `path`.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            if (!didFindAccounts) break;
+            if (onIntermediateResult) onIntermediateResult(foundAccounts);
+        }
+
+        // check for existing labels
+        const walletId = await deriveWalletId(foundAccounts);
+        const walletInfo = await WalletStore.Instance.get(walletId);
+        const existingAccounts: Map<string, AccountInfo> = walletInfo ? walletInfo.accounts : new Map();
```
sisou: As discussed shortly, I don't feel comfortable having WalletStore code in this library. Since for a Keyguard result we already have this info and this is duplicate effort. I think it should be handled in the user of his library after this library returns with all found account addresses and paths.

@nibhar ?

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+            const balances = await NetworkClient.Instance.getBalance(userFriendlyAddresses);
+            for (const account of derivedAccounts) {
+                const balance = balances.get(account.address);
+                if (balance === undefined || balance === 0) continue;
+                foundAccounts.push(Object.assign({}, account, { balance }));
+                didFindAccounts = true;
+            }
+
+            if (!didFindAccounts) break;
+            if (onIntermediateResult) onIntermediateResult(foundAccounts);
```
sisou: Can you explain shortly what this hook can be used for?

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        // check for existing labels
+        const walletId = await deriveWalletId(foundAccounts);
+        const walletInfo = await WalletStore.Instance.get(walletId);
+        const existingAccounts: Map<string, AccountInfo> = walletInfo ? walletInfo.accounts : new Map();
+        return foundAccounts.map((account) => new AccountInfo(
+            account.keyPath,
```
sisou: .

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+            // find accounts with a balance > 0
+            // TODO should use transaction receipts combined with balance
+            let didFindAccounts = false;
+            const userFriendlyAddresses = derivedAccounts.map(({ address }) => address);
+            const balances = await NetworkClient.Instance.getBalance(userFriendlyAddresses);
+            for (const account of derivedAccounts) {
+                const balance = balances.get(account.address);
+                if (balance === undefined || balance === 0) continue;
+                foundAccounts.push(Object.assign({}, account, { balance }));
+                didFindAccounts = true;
```
sisou: The `didFindAccounts` condition actually doesn't honor the `MAX_ALLOWED_GAP` constant, as it makes the lib potentially search up to 39 accounts ahead.

Like so: if in any block of 20 accounts searched, only the very first is detected as active, `didFindAccounts` is true and the loop runs another time, searching 39 accounts further than the last active index.

That's why my previous implementation had both the `nextIndexToDerive` and the `lastActiveIndex`, to enable giving the call to `deriveAccounts` only the relevant number of remaining indices inside the gap.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            // already start deriving next accounts
+            derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+            nextStartIndex += derivedAccounts.length; // by always advancing in blocks of MAX_ALLOWED_GAP it can happen
+            // that we include accounts that are more than MAX_ALLOWED_GAP apart (e.g. one towards the beginning of
+            // the first block, the other one towards the end of the second block). But that's actually not bad.
+
+            // find accounts with a balance > 0
+            // TODO should use transaction receipts combined with balance
```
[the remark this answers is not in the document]
sisou: It's a bit more involved than that, because you are potentially facing rate-limiting by peers when requesting `transactionHistory`, thus it would need additional fail-safes, which are outside the scope of this PR.

I created https://github.com/nimiq/accounts/issues/94 to track it.

src/router.ts
```diff
@@ -18,12 +20,11 @@ const Login                   = () => import(/* webpackChunkName: "login" */ './
 const LoginSuccess            = () => import(/* webpackChunkName: "login" */ './views/LoginSuccess.vue');
 const LoginErrorHandler       = () => import(/* webpackChunkName: "login" */ './views/LoginErrorHandler.vue');
 
-// const ExportFile              = () => import(/* webpackChunkName: "export" */ './views/ExportFile.vue');
-// const ExportWords             = () => import(/* webpackChunkName: "export" */ './views/ExportWords.vue');
+// const ExportFile           = () => import(/* webpackChunkName: "export" */ './views/ExportFile.vue');
+// const ExportWords          = () => import(/* webpackChunkName: "export" */ './views/ExportWords.vue');
```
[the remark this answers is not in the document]
sisou: Tracked in https://github.com/nimiq/accounts/issues/86.

src/views/LoginSuccess.vue
sisou: Tracked in https://github.com/nimiq/accounts/issues/86.

src/lib/AccountFinder.ts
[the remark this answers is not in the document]
danimoh: The nano-api currently supports a method `requestTransactionHistory`, ideally we'd however only ask for transaction receipts, which is much cheaper to calculate.
On the other hand of course one doesn't import accounts that often... So for now `requestTransactionHistory` could be ok.

No, I'd say accounts that have transactions OR a balance. An account that has a balance without (incoming) transactions is a rare corner case though. It would be either a genesis account (these are all legacy wallets though for which we don't have to run the account finding anyways) or accounts that mined a block. So we could also drop the ""OR a balance" part to make the call cheaper.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+        deriveWalletId: (accounts: Array<{ address: string, keyPath: string, balance: number }>) => Promise<string>,
+        defaultLabel: string = 'Standard Account',
+        onIntermediateResult?: (accounts: Array<{ address: string, keyPath: string, balance: number }>) => void,
+    ): Promise<AccountInfo[]> {
+        await NetworkClient.Instance.init(); // initialize NetworkClient if not initialized yet
+        const foundAccounts: Array<{ address: string, keyPath: string, balance: number }> = [];
+        let nextStartIndex = 0;
+        let derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+        nextStartIndex += AccountFinder.MAX_ALLOWED_GAP;
+        while (true) {
```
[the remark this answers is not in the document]
danimoh: yes, nice

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+            const derivedAccounts = await derivedAccountsPromise;
+            if (derivedAccounts.length !== AccountFinder.MAX_ALLOWED_GAP) {
+                throw new Error('\'deriveAccounts\' returned wrong number of accounts');
+            }
+
+            // already start deriving next accounts
+            derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+            nextStartIndex += derivedAccounts.length; // by always advancing in blocks of MAX_ALLOWED_GAP it can happen
```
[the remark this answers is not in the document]
danimoh: Yes, `derivedAccounts.length === Max_ALLOWED_GAP` always holds true.
Maybe writing MAX_ALLOWED_GAP is more readable.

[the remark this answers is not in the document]
nibhar: But that only works iff `derivedAccounts.length === Max_ALLOWED_GAP`.

src/views/LoginSuccess.vue
```diff
@@ -61,142 +63,116 @@ export default class LoginSuccess extends Vue {
...
+                this.accounts = currentWalletInfo.accounts;
+            }
         }
 
-        // 2. Derive accounts from paths
+        // The Keyguard always returns (at least) one derived Address,
+        // which we'll also already add and display
+        const keyguardResultAccounts = this.keyguardResult.addresses.map((addressObj) => ({
+            address: new Nimiq.Address(addressObj.address).toUserFriendlyAddress(),
+            keyPath: addressObj.keyPath,
```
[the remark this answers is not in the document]
danimoh: It's just a code simplification. `_addAccounts` is generalized to potentially update the balance which does not apply here, but still just calling the same method makes the code shorter and more readable.

I'm using keyPath, because it's called like this in the KeyguardResult and therefore also in the LedgerApi result.

src/views/LoginSuccess.vue
```diff
@@ -61,142 +63,116 @@ export default class LoginSuccess extends Vue {
...
+                keyPath = account.path;
+            } else {
+                userFriendlyAddress = account.address;
+                keyPath = account.keyPath;
+            }
 
-        // 3. Get balances for the derived accounts
+            const updatedAccountInfo = this.accounts.get(userFriendlyAddress)
+                || new AccountInfo(keyPath, LoginSuccess.DEFAULT_ACCOUNT_LABEL,
+                    Nimiq.Address.fromUserFriendlyAddress(userFriendlyAddress));
```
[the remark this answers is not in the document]
danimoh: I changed the line-breaks now.

Regarding the setting of accounts: It's just shorter code and setting the accounts is a cheap operation.
But although I find that it's not necessary, I added a check now.

src/views/LoginSuccess.vue
```diff
@@ -61,142 +63,116 @@ export default class LoginSuccess extends Vue {
...
+    }
 
-        // FIXME: Use LedgerClient here for LEDGER keys
-        const serializedAddresses = await this.keyguard.deriveAddresses(
-            this.keyguardResult.keyId,
-            pathsToDerive.slice(this.lastDerivedIndex + 1), // We don't want to send `empty` paths into the Keyguard
-        );
-        const userFriendlyAddresses = serializedAddresses
-            .map((rawAddr) => new Nimiq.Address(rawAddr).toUserFriendlyAddress());
+    private _addAccounts(accounts: Array<{ address: string, keyPath: string, balance?: number} | AccountInfo>) {
```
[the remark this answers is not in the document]
danimoh: I don't like creating an AccountInfo with an empty label. Cleaner would be to make the label in AccountInfo optional. But I also don't know whether we want to do that.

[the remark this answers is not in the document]
sisou: We can introduce another internal type that has is like `AccountInfo` but without label.

src/lib/AccountFinder.ts
danimoh: They are bot the same.
This comment is the answer for your remark at line 38.

src/lib/AccountFinder.ts
danimoh: It's an unexpected error that only get's thrown if we messed up in the code.
The `deriveAccounts` method knows via a parameter how may accounts it should derive.
This check could theoretically also be removed.

src/lib/AccountFinder.ts
danimoh: yes, it is

src/lib/AccountFinder.ts
danimoh: Mostly for UI updates:
1 account found...
3 accounts found...
5 accounts found...
etc.

src/lib/AccountFinder.ts
danimoh: Will probably keep this behavior, as discussed.

src/views/LoginSuccess.vue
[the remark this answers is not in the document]
sisou: Ok, so the hardware and software wallets return it, fine, but in the Accounts Manager we use `path`.

src/lib/AccountFinder.ts
sisou: Then at least improve the comment for now, that it's not only `one towards the end` but up to 39 (or `more towards to the end`, and replace `that's not bad` with why it's not bad.

src/lib/AccountFinder.ts
sisou: Ok, then if we see this in Sentry, we know it broke a users login flow.

sisou: On the other hand, does this actually break anything if less accounts are returned? I don't think so, because you are using the `derivedAccounts.length` to advance the index anyway. In that case you can just remove the check and continue regularily.

src/lib/AccountFinder.ts
sisou: It is still marked as *experimental* [on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Browser_compatibility). It is widely supported, but since we are not using it anywhere else and you are integrating with the existing code, this should use standard assignment instead.

src/views/LoginSuccess.vue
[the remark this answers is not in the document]
danimoh: The situation now has improved. `foundAccounts` creates real `AccountInfo` instances and also `onIntermediateResult` takes an `AccountInfo` array now.

`deriveAccounts` and `deriveWalletId` still take `{ address, keyPath }` objects though.
But this is only visible in the method signature and also sort of serves as a documentation for the expected arguments and return types, where you can directly see which properties the objects should contain.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
+import { AccountInfo } from '@/lib/AccountInfo';
+import { NetworkClient } from '@nimiq/network-client';
+import { WalletStore } from '@/lib/WalletStore';
+
+export default class AccountFinder {
+    public static async findAccounts(
+        deriveAccounts: (startIndex: number, count: number) => Promise<Array<{ address: string, keyPath: string }>>,
+        deriveWalletId: (accounts: Array<{ address: string, keyPath: string, balance: number }>) => Promise<string>,
+        defaultLabel: string = 'Standard Account',
+        onIntermediateResult?: (accounts: Array<{ address: string, keyPath: string, balance: number }>) => void,
```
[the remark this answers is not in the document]
danimoh: this one is using AccountInfo now

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,58 @@
...
+    ): Promise<AccountInfo[]> {
+        await NetworkClient.Instance.init(); // initialize NetworkClient if not initialized yet
+        const foundAccounts: Array<{ address: string, keyPath: string, balance: number }> = [];
```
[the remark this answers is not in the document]
danimoh: also using `AccountInfo`

src/lib/AccountFinder.ts
sisou: All right, let's see if it comes up in the testnet.

src/lib/AccountFinder.ts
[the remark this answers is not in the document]
danimoh: Ok, so you're suggesting to replace `derivedAccounts.length` by `AccountFinder.MAX_ALLOWED_GAP` in the while loop? Or what is it about?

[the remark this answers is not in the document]
sisou: Yes, he is suggesting to replace it in this line, since the two should be equal always.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,66 @@
...
+            // already start deriving next accounts
+            derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+            nextStartIndex += derivedAccounts.length; // by always advancing in blocks of MAX_ALLOWED_GAP it can happen
+            // that we include accounts that are more than MAX_ALLOWED_GAP apart (e.g. one account at index 0, the other
+            // one at index 39). But while this does not exactly follow the specification, it's actually good for the
+            // user, as we'll potentially find more of the user's accounts. For example, if the user adds the accounts
+            // derived with indices 0, 19, 39 to his wallet but then only ends up using accounts 0 and 39, the account
+            // at index 19 will not be found anymore on reimport. With the current implementation however, at least the
+            // account 39 would be found, while an implementation strictly following the specification would stop the
+            // search at index 20.
```
sisou: Here is my suggestion for the wording of this comment:

> derivedAccounts.length === MAX_ALLOWED_GAP because of the check above
> By always advancing in groups of MAX_ALLOWED_GAP addresses per round, it often happens that more addresses are derived and checked for activity than the BIP44 address gap limit (https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit) stipulates, because whenever an active address in a group of addresses is found, the next full group is also derived. Thus the actual gap limit of this implementation is (2 x MAX_ALLOWED_GAP) - 1.
> We argue that this is good UX for users, as potentially more of their active addresses are found, even if they haven't strictly followed to the standard - at only a relatively small cost to the network.

src/lib/AccountFinder.ts
```diff
@@ -0,0 +1,66 @@
...
+        const foundAccounts: AccountInfo[] = [];
+        let nextStartIndex = 0;
+        let derivedAccountsPromise = deriveAccounts(nextStartIndex, AccountFinder.MAX_ALLOWED_GAP);
+        nextStartIndex += AccountFinder.MAX_ALLOWED_GAP;
+
+        // calculate walletId (potentially using the first account) and read existing accounts
+        const existingAccountsPromise = derivedAccountsPromise
+            .then((derivedAccounts) => deriveWalletId(derivedAccounts[0]))
+            .then((walletId) => WalletStore.Instance.get(walletId))
+            .then((walletInfo) => walletInfo ? walletInfo.accounts : new Map<string, AccountInfo>());
```
sisou: I still think this can be a callback to the user of this library, which would save us a database access for all Keyguard users.

src/views/LoginSuccess.vue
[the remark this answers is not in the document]
danimoh: Quickly discussed this with Sebastian. He thinks having plain object types without an additional type definition is suitable here.

src/lib/AccountFinder.ts
danimoh: Sebastian brought up the idea that the semantics of the `AccountFinder` could be changed to a `WalletImporter` (better name still needs to be found) that handles everything that's related to accound finding + WalletInfo creation. What do you think?

src/views/LoginSuccess.vue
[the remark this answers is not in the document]
danimoh: Using an internal type definition `BasicAccountInfo` within `WalletInfoCollector` now.

src/lib/AccountFinder.ts
danimoh: indeed. This code has now also changed though

src/lib/WalletInfoCollector.ts
```diff
@@ -0,0 +1,214 @@
...
+                WalletInfoCollector.MAX_ALLOWED_GAP, walletType, walletId);
+
+            // find accounts with a balance > 0
+            didFindAccounts = false;
+            // TODO should use transaction receipts
+            const balances = await WalletInfoCollector._getBalances(derivedAccounts!);
+            const usedAccounts: BasicAccountInfo[] = [];
+            for (const account of derivedAccounts!) {
+                const balance = balances.get(account.address);
+                if (balance === undefined || balance === 0) continue;
```
sisou: Could use `if (!balance)` maybe?

src/lib/WalletInfoCollector.ts
```diff
@@ -0,0 +1,214 @@
...
+                usedAccounts.push(account);
+                didFindAccounts = true;
+            }
+            if (didFindAccounts) {
```
sisou: Instead of `didFindAccounts`, you could use `usedAccounts.length > 0` and would get two flys with one stone ;)

src/views/LoginSuccess.vue
```diff
@@ -30,226 +29,108 @@ import { Component, Emit, Vue } from 'vue-property-decorator';
...
-        // can be displayed while waiting for the network.
-        // Before adding the WalletInfo object, make sure it is not set already.
-        this.keyguardResult.addresses.forEach((addressObj) => {
-            const address = new Nimiq.Address(addressObj.address);
-            if (!this.accounts.has(address.toUserFriendlyAddress())) {
-                const addressInfo = new AccountInfo(
-                    addressObj.keyPath,
-                    this.defaultAccountLabel,
-                    address,
+        const keyguardResultAccounts = this.keyguardResult.addresses.map((addressObj) => ({
```
sisou: I'm guessing this should be a `BasicAccountInfo[]` type? Could you add the annotation for it?

src/views/LoginSuccess.vue
```diff
@@ -30,226 +29,108 @@ import { Component, Emit, Vue } from 'vue-property-decorator';
...
+            address: new Nimiq.Address(addressObj.address).toUserFriendlyAddress(),
+            path: addressObj.keyPath,
+        }));
+
+        let tryCount = 0;
+        while (true) {
+            try {
+                ++tryCount;
```
sisou: I think in the Keyguard we have a linting rule that forbids `++` outside of for loops. So this would be clearer as `tryCount += 1;`.

src/views/LoginSuccess.vue
```diff
@@ -30,226 +29,108 @@ import { Component, Emit, Vue } from 'vue-property-decorator';
...
+                const walletInfo = await WalletInfoCollector.collectWalletInfo(
+                    this.keyguardResult.keyType,
+                    this.keyguardResult.keyId,
+                    keyguardResultAccounts,
+                    (updatedWalletInfo) => this._onWalletInfoUpdate(updatedWalletInfo),
                 );
-                this.accounts.set(addressInfo.userFriendlyAddress, addressInfo);
+                this._onWalletInfoUpdate(walletInfo);
```
sisou: `_onWalletInfoUpdate` is already called in each `WalletInfoCollector._addAccounts`, also after a balance update, so calling it again here after finishing the WalletInfo retrieval is not necessary.

src/views/LoginSuccess.vue
```diff
@@ -30,226 +29,108 @@ import { Component, Emit, Vue } from 'vue-property-decorator';
...
+                this.retrievalFailed = false;
+                break;
+            } catch (e) {
+                this.retrievalFailed = true;
+                if (tryCount === 5) throw e;
```
sisou: I always like `>= 5` better, just in case the loop messes up somehow and skips 5.

src/views/LoginSuccess.vue
```diff
@@ -293,8 +174,8 @@ export default class LoginSuccess extends Vue {
         margin-right: 1rem;
     }
 
-    .network-wrapper {
-        padding: 1rem;
+    .loader {
+        min-height: 168px;
```
sisou: In `rem`, please.

src/lib/WalletInfoCollector.ts
sisou: On second thought, this is actually better in this instance as is.

src/lib/WalletInfoCollector.ts
```diff
@@ -0,0 +1,214 @@
+import { NetworkClient } from '@nimiq/network-client';
+import { KeyguardClient } from '@nimiq/keyguard-client';
+import { AccountInfo } from '@/lib/AccountInfo';
+import { WalletStore } from '@/lib/WalletStore';
+import { WalletInfo, WalletType } from '@/lib/WalletInfo';
+import LazyLoading from '@/lib/LazyLoading';
+
+type BasicAccountInfo = { // tslint:disable-line:interface-over-type-literal
```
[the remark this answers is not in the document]
danimoh: you guys said that you prefer type definitions ...

src/views/LoginSuccess.vue
danimoh: It is a `BasicAccountInfo[]` yes, but that type is currently only available internally in `WalletInfoCollector` and I'm not much in favor of exporting it as exporting some sort of `AccountInfo` from a `WalletInfoCollector` sounds weird to me. It would rather be something like `WalletInfoCollectorIntermediateAccount` which is not a snappy name. Also I don't like having several account info types that are about the same.
I think keeping it the way it is best. If the types don't match, TypeScript is already gonna recognize this.

src/views/LoginSuccess.vue
danimoh: Yes, but that's an iternal code detail that is not visible from the outside.

src/lib/WalletInfoCollector.ts
```diff
@@ -0,0 +1,214 @@
...
+    private static _wasmInitializationPromise?: Promise<void>;
+
+    private static _initializeDependencies(walletType: WalletType) {
+        WalletInfoCollector._networkInitializationPromise =
+            WalletInfoCollector._networkInitializationPromise || NetworkClient.Instance.init();
+        if (walletType === WalletType.BIP39) {
+            WalletInfoCollector._keyguardClient = WalletInfoCollector._keyguardClient || new KeyguardClient();
+        } else if (walletType === WalletType.LEDGER) {
+            WalletInfoCollector._wasmInitializationPromise =
+                WalletInfoCollector._wasmInitializationPromise || LazyLoading.loadNimiq(true);
```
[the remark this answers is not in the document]
danimoh: I think `await Nimiq.CryptoWorker.getInstanceAsync();` is not enough to load the wasm module. I'm just using plain `WasmHelper.doImportBrowser()` now and opened a PR in the core to address the shortcomings of the `doImportBrowser` method without the need to build around it.

src/views/LoginSuccess.vue
sisou: But you explicitly give the function to the WalletCollector and thus know that it's called on each change.

danimoh: I renamed the parameter `onIntermediateResult` to `onUpdate` to make it more explicit that the api can be used this way.

src/lib/WalletInfoCollector.ts
```diff
@@ -0,0 +1,217 @@
...
+            new Map<string, AccountInfo>(),
+            [],
+            walletType,
+            false,
+        );
+    }
+
+    private static async _deriveAccounts(startIndex: number, count: number, walletType: WalletType, walletId?: string)
+        : Promise<BasicAccountInfo[]> {
+        if (walletType === WalletType.LEGACY) {
```
nibhar: This is 4 cases for a variable check.
At that point i usually change it to a switch statement, even more so if every case either returns or throws.

Again, this is only personal preference. I am interested in your opinion mostly.

src/views/LoginSuccess.vue
```diff
@@ -30,226 +29,107 @@ import { Component, Emit, Vue } from 'vue-property-decorator';
...
-                const addressInfo = this.accounts.get(addr);
-                addressInfo!.balance = Nimiq.Policy.coinsToSatoshis(balances.get(addr) || 0);
-                this.accounts.set(addr, addressInfo!);
-            });
-            this.accountsUpdateCount += 1;
-
-            // Kick off account detection
-            this.findAccounts();
-        }
+        // TODO network visuals with longer than 1 list of accounts during retrieval
```
nibhar: As far as I know this TODO is obsolete, as the screen will not show the account list.

danimoh: I actually do not know what is meant by this TODO

sisou: Will be removed when that flow is reworked.

src/lib/WalletInfoCollector.ts
sisou: Yes, I agree a switch here would look better.

PR #92, opened 2018-12-07
Use Loader component for migrate request, add return indicator to Loader

sisou:
The Loader redirect indicator is displayed at the bottom of the screen and fills clockwise in 1.5 seconds ('ease' animation):
image

### DIFF
--- a/src/components/Loader.vue
+++ b/src/components/Loader.vue
@@ -111,6 +111,7 @@ class Loader extends Vue {
     public static readonly SUCCESS_REDIRECT_DELAY: number = 2000; // 1s of transition + 1s of display
 
     @Prop({type: String, default: 'Improving the world'}) private title!: string;
+    // Using Loader.State.LOADING here results in runtime error: 'Cannot read property 'LOADING' of undefined'
     @Prop({type: String, default: 'loading'}) private state!: string;
     @Prop(Boolean) private lightBlue?: boolean;
     @Prop(String) private status?: string;
--- a/src/components/Loader.vue
+++ b/src/components/Loader.vue
@@ -121,7 +121,10 @@ class Loader extends Vue {
     private currentStatus: string = '';
     private nextStatus: string = '';
     private isStatusTransitioning: boolean = false;
-    private strokeDashoffset: number = 14 * 2 * Math.PI;
+
+    // Stroke offset used to animate SVG redirect indicator from full-offset to no-offset
+    private STROKE_DASHOFFSET: number = 14 * 2 * Math.PI;
+    private strokeDashoffset: number = this.STROKE_DASHOFFSET;
 
     /**
      * To enable a smooth transition of the non-transitionable background-image
@@ -142,6 +145,7 @@ class Loader extends Vue {
 
     // In a browser, setTimout returns a number, but Typescript thinks it should be a NodeJS.Timeout
     private stateUpdateTimeout: any = null;
+    private indicatorDisplayTimeout: any = null;
 
     @Watch('state', {immediate: true})
     private updateState(newState: string, oldState: string) {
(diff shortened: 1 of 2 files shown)


src/components/Loader.vue
```diff
@@ -99,8 +107,12 @@ import { Component, Prop, Watch, Emit, Vue } from 'vue-property-decorator';
  */
 @Component
 class Loader extends Vue {
+    // TODO: Move to CONSTANTS
+    public static readonly SUCCESS_REDIRECT_DELAY: number = 2000; // 1s of transition + 1s of display
+
     @Prop({type: String, default: 'Improving the world'}) private title!: string;
-    @Prop({type: String, default: Loader.Status.LOADING}) private state!: string;
+    // Using Loader.State.LOADING here results in runtime error: 'Cannot read property 'LOADING' of undefined'
+    @Prop({type: String, default: 'loading'}) private state!: string;
```
danimoh: could be typed more strictly:
```ts
 @Prop({ default: 'loading' as Loader.State }) private state!: Loader.State;
```

src/components/Loader.vue
```diff
@@ -130,25 +146,35 @@ class Loader extends Vue {
 
     // In a browser, setTimout returns a number, but Typescript thinks it should be a NodeJS.Timeout
     private stateUpdateTimeout: any = null;
+    private indicatorDisplayTimeout: any = null;
```
danimoh: when using `window.setTimeout`, it knows that you are using the browser verion. Can thus be typed as number.

src/components/Loader.vue
```diff
@@ -111,6 +123,10 @@ class Loader extends Vue {
     private nextStatus: string = '';
     private isStatusTransitioning: boolean = false;
 
+    // Stroke offset used to animate SVG redirect indicator from full-offset to no-offset
+    private STROKE_DASHOFFSET: number = 14 * 2 * Math.PI;
```
danimoh: should be `static readonly`

src/views/Migrate.vue
```diff
@@ -10,33 +18,56 @@ import { Component, Vue } from 'vue-property-decorator';
...
+    private message: string = '';
+
+    private static: {keyguardClient?: KeyguardClient} = {};
+
+    public mounted() {
+        this.static.keyguardClient = this.$rpc.createKeyguardClient();
+
+        // Need to freeze the object to prevent Vue's reactivity system
+        // from accessing properties of the cross-origin `window.opener`,
+        // which throws security errors in browsers.
```
danimoh: Why not actually declare the keyguardClient as static property of the class to avoid this problem?
Naming it `_keyguardClient` might also work.

src/views/Migrate.vue
```diff
@@ -52,16 +83,36 @@ export default class Migrate extends Vue {
...
-        this.status = 'Done.';
+        this.title = 'Migration completed.';
+        this.state = Loader.State.SUCCESS;
         const walletInfoEntries = walletInfos.map((walletInfo) => walletInfo.toObject());
-        setTimeout(() => this.$rpc.resolve(walletInfoEntries), Migrate.RETURN_WAIT_TIME);
+        setTimeout(() => this.$rpc.resolve(walletInfoEntries), Loader.SUCCESS_REDIRECT_DELAY);
+    }
+
+    private onError(error: Error) {
+        this.title = 'Ups, something went wrong';
```
danimoh: Is it also "ups" in English? Or rather "whoops" or something else?

src/views/Migrate.vue
nibhar: maybe even 
```
private static: {keyguardClient?: KeyguardClient};
``` 
does the trick. If i recall correctly, uninitialised properties will not be made reactive by vue.

src/views/Migrate.vue
```diff
@@ -52,16 +83,36 @@ export default class Migrate extends Vue {
...
+        this.message = `${error.name}: ${error.message}`;
+        this.state = Loader.State.ERROR;
+        throw error; // Notify sentry
+    }
+
+    private tryAgain() {
```
nibhar: Is there good reason to assume, that simply waiting one second will change the outcome of this request?

sisou: I don't know, but as discussed, simply returning the user does not help either, because without the migration the Safe is non-functional. So we better not see this error in Sentry...

src/components/Loader.vue
sisou: Cool!

src/views/Migrate.vue
sisou: @nibhar You, sir, are a genius! :nerd_face:

PR #95, opened 2018-12-08
Add new optional checkout parameter 'validityDuration'

### LINKED ISSUE #84: Merchants should be able to set tx validity length
Exchanges mostly only guarantee an exchange rate for a certain amount of time. Thus a merchant should be able to set for how many blocks (~minutes) a transaction should be valid.

We would then use this to offset the `validityStartHeight` of the transaction.

sisou:
Resolves #84.

Also starts a `Constants.ts` (#86).

### DIFF
--- a/src/views/CheckoutOverview.vue
+++ b/src/views/CheckoutOverview.vue
@@ -23,6 +23,8 @@ import Network from '../components/Network.vue';
 
 @Component({components: {Amount, PageFooter, CheckoutDetails, Network}})
 export default class CheckoutOverview extends Vue {
+    private static readonly TX_VALIDITY_WINDOW: number = 120;
+
     @Static private rpcState!: RpcState;
     @Static private request!: ParsedCheckoutRequest;
 
@@ -48,6 +50,11 @@ export default class CheckoutOverview extends Vue {
             return;
         }
 
+        const validityStartHeight = this.height
+            + 1
+            - CheckoutOverview.TX_VALIDITY_WINDOW
+            + this.request.validityDuration;
+
         const request: KSignTransactionRequest = {
             layout: 'checkout',
             shopOrigin: this.rpcState.origin,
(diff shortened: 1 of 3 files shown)


src/views/CheckoutOverview.vue
```diff
@@ -23,6 +23,8 @@ import Network from '../components/Network.vue';
 
 @Component({components: {Amount, PageFooter, CheckoutDetails, Network}})
 export default class CheckoutOverview extends Vue {
+    private static readonly TX_VALIDITY_WINDOW: number = 120;
```
nibhar: that looks like a re-definition to me.

src/views/CheckoutOverview.vue
```diff
@@ -48,6 +50,11 @@ export default class CheckoutOverview extends Vue {
             return;
         }
 
+        const validityStartHeight = this.height
+            + 1
+            - CheckoutOverview.TX_VALIDITY_WINDOW
```
nibhar: should use the constant from constants.ts

src/views/CheckoutOverview.vue
```diff
@@ -48,6 +50,11 @@ export default class CheckoutOverview extends Vue {
             return;
         }
 
+        const validityStartHeight = this.height
+            + 1
```
nibhar: where does the 1 come from?

src/views/CheckoutOverview.vue
```diff
@@ -93,7 +99,7 @@ export default class CheckoutOverview extends Vue {
 
     @Watch('height')
     private logHeightChange(height: number, oldHeight: number) {
-        console.log(`Height changed from ${oldHeight} to ${height}`);
+        console.debug(`Got height: ${height} (was ${oldHeight})`);
```
nibhar: is this needed? aside from helping us developing it, this watcher probably should be removed prior to release.

PR #102, opened 2018-12-11
Add label length restriction

### LINKED ISSUE #80: Restrict label length
Both wallet and account labels should be restricted to 63 bytes (not 64, to be able to store it in 6 bits while keeping the possibility of having a 0 length (which would then correspond to default labels)).

**Note:** 63 bytes does not mean 63 characters, as non-ascii characters can be multi-byte. Javascript's `String.length` is also not usable, as JS uses a 16-bit encoding, not 8-bit, and may thus report a shorter string length than would be in bytes.

sisou:
Via a custom label input component.

Resolves #80.

### DIFF
--- a/src/components/LabelInput.vue
+++ b/src/components/LabelInput.vue
@@ -1,0 +1,10 @@
+<script lang="ts">
+    import { Component, Prop, Vue } from 'vue-property-decorator';
+    import { LabelInput as StandardLabelInput } from '@nimiq/vue-components';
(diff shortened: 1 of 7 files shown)


Bettelstab:
I wanted to wait with the PR until https://github.com/nimiq/vue-components/pull/4 is merged and there is a new version number for vue-components, but I guess it will find it's way eventually. However, this PR alone doesn't resolve #80 ;)

sisou:
Reverted `master` to before the merge

Issue #103, opened 2018-12-12
Loading spinner doesn't work in built version


Bettelstab:
Somehow the data url is different from the source.

Bettelstab:
I remember investigating this before and it may be related to some bug in the interplay of vue-loader and postcss.

Temporarily resolved by https://github.com/nimiq/accounts/pull/104

sisou:
Does #104 also resolve the issue in the Loader component? Or should we remove the CSS for the loading animation there and use the global definition of `loading.css`?

sisou:
Fixed by #104 and new LoadingSpinner component in @nimiq/vue-components.

PR #104, opened 2018-12-12
Fast first render

### LINKED ISSUE #71: Improve loading time / time until first render
First visible render is only after Vue and essentially the whole app has loaded, which takes a long time on slower internet connection. The loading spinner should already be visible instantly.

Bettelstab:
Displays the loading screen while loading the vue app.

Resolves #71, #103.


public/loading.css
```diff
@@ -0,0 +1,87 @@
...
+    outline: none !important;
+}
+
+.logo .nimiq-logo {
+    height: 4rem;
+    width: 4rem;
+    margin-right: 2rem;
+}
+
+.nq-icon {
```
sisou: When you move the `.nq-*` rules further up, you can maybe reduce some of the class-specificity for the logo.

sisou:
I think the CSS for the header, logo and Loader can be removed from `App.vue` then, because it's already in the `loading.css` file registered globally, right?

sisou:
Also, you accidentally added the `demos/nimiq.png` file with this PR.

public/index.html
```diff
@@ -10,12 +11,28 @@
 
     <link href="https://fonts.googleapis.com/css?family=Muli:400,600,700" rel="stylesheet">
     <link href="https://fonts.googleapis.com/css?family=Fira+Mono&text=0123456789ABCDEFGHJKLMNPQRSTUVXY" rel="stylesheet">
-  </head>
-  <body>
+    <link href="loading.css" rel="stylesheet">
```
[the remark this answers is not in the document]
Bettelstab: I'll go ahead with those changes, but the last one. I made it this way with CSP (disallowing inline styles) in mind. What do you think?

public/loading.css
Bettelstab: I don't really get it.

public/index.html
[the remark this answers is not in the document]
Bettelstab: With the fonts, it so no easy: https://developers.google.com/web/updates/2016/02/font-display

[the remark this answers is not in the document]
danimoh: Yeah, ok, I also don't know whether it would actually be beneficial to inline the style. By not inlining, it can at least download stuff in parallel.

Loading the fonts is done by linking to the google stylesheets though and the download of the stylesheets is render blocking even though the download of the font itself is not. Thus, we should move these `<link rel="stylesheet">` tags to the end of `<body>` or maybe just include the content of the google font stylesheet into our stylesheet. We'd need to be careful to include the right fallbacks though, as google serves different stylesheets depending on the user's browser.
So, performance wise, including that stuff into our stylesheet might actually be best.

public/blocking.css
```diff
@@ -97,4 +97,10 @@ body {
 
 .nq-icon.nimiq-logo {
     background-image: url('data:image/svg+xml,<svg width="27" height="24" viewBox="0 0 27 24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M26.6991 10.875L21.0741 1.125C20.6691 0.4275 19.9266 0 19.1241 0H7.87414C7.07164 0 6.32914 0.4275 5.92789 1.125L0.302891 10.875C-0.0983594 11.5725 -0.0983594 12.4275 0.302891 13.125L5.92789 22.875C6.32914 23.5725 7.07164 24 7.87414 24H19.1241C19.9266 24 20.6691 23.5725 21.0704 22.875L26.6954 13.125C27.1004 12.4275 27.1004 11.5725 26.6991 10.875Z" fill="url(%23paint0_radial)"/><defs><radialGradient id="paint0_radial" cx="0" cy="0" r="1" gradientUnits="userSpaceOnUse" gradientTransform="translate(26.9996 24) rotate(-180) scale(26.9977 24)"><stop stop-color="%23EC991C"/><stop offset="1" stop-color="%23E9B213"/></radialGradient></defs></svg>');
+}
+
+@media (max-width: 450px) {
+    html {
+        font-size: 7px;
```
danimoh: Need to set `--nimiq-size` to 7px then also. Actually it would be best to just do
```css
--nimiq-size: 1rem;
```
Then it doesn't need to be changed every time we use another root font size.

src/App.vue
```diff
@@ -56,13 +56,8 @@ export default class App extends Vue {
 </script>
 
 <style>
-    html, body {
-        height: 100%;
-    }
-
     @media (max-width: 450px) {
         html {
-            font-size: 7px;
```
danimoh: remove the `--nimiq-size` here

public/index.html
```diff
@@ -5,12 +5,9 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <script src="https://cdn.nimiq-testnet.com/web-offline.js"></script>
+    <script src="https://cdn.nimiq-testnet.com/web-offline.js" defer></script>
```
danimoh: Is there a way to add `defer` to the `index.js` that webpack injects too?
Otherwise, I think, `index.js` will be executed before `web-offline.js`. Thus, if there is already a call to `Nimiq.something` at that point, it would fail.

public/loading.css
```diff
@@ -1,10 +1,23 @@
+html, body {
+    height: 100%;
```
danimoh: maybe better use min-height here also?

public/index.html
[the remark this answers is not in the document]
danimoh: BTW, `<div style="flex-grow: 1;"></div>` will also not work with CSP in place. Also see my other comment for an alternative approach.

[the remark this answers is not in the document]
Bettelstab: The latter: I know, already replaced it, just forgot to push.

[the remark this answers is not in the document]
Bettelstab: The former: we could use a self-hosted version of webfontloader, or program the functionality ourselves. But I doubt it's worth the effort for 3 seconds without fonts (takes a lot longer to load with slow 3G)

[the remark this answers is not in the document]
danimoh: Yeah, hosting that ourselves doesn't make much sense.
I was really just talking about copying the content of `https://fonts.googleapis.com/css?family=Muli:400,600,700` and `https://fonts.googleapis.com/css?family=Fira+Mono&text=0123456789ABCDEFGHJKLMNPQRSTUVXY` into our css.
And maybe making sure that links to woff (google hosted) as a fallback are also included.
But probably even woff2 is enough. Browser support is pretty good.

public/index.html
sisou: Yes, I'm running into this when merging this. So I suggest moving the Nimiq lib to the bottom directly above the build injection comment and removing the defer.

danimoh: But by using `defer` in the head the browser immediately starts downloading the file.

sisou: Then we need to make the build js files also have defer. Maybe this can be integrated: https://www.npmjs.com/package/script-ext-html-webpack-plugin

public/index.html
```diff
@@ -29,8 +29,8 @@ <h2>Loading, hold on</h2>
         </div>
         <div style="flex-grow: 1;"></div>
     </div>
-    <link href="https://fonts.googleapis.com/css?family=Muli:400,600,700" rel="stylesheet">
-    <link href="https://fonts.googleapis.com/css?family=Fira+Mono&text=0123456789ABCDEFGHJKLMNPQRSTUVXY" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css?family=Muli:400,600,700" rel="preload" as="style" onload="this.rel='stylesheet'">
+    <link href="https://fonts.googleapis.com/css?family=Fira+Mono&text=0123456789ABCDEFGHJKLMNPQRSTUVXY" rel="preload" as="style" onload="this.rel='stylesheet'">
```
danimoh: Cool but probably not CSP conform.
Also will trigger the download quite late, i.e. when the browser is idle and probably after the other preloads that webpack injects into the head.

sisou: Webpack is using `prefetch`. `preload` downloads directly, I think.

public/index.html
Bettelstab: That plugin didn't work (html outcome ignored our index.html, maybe needed more configuration...). I wrote a tiny script instead to do that.

public/loading.css
Bettelstab: Yep, seems to be good.

Bettelstab:
Please check again @danimoh @sisou

public/index.html
```diff
@@ -5,10 +5,12 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <script src="https://cdn.nimiq-testnet.com/web-offline.js" defer></script>
+    <script src="https://cdn.nimiq-testnet.com/web-offline.js"></script>
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title>Nimiq Accounts Manager</title>
     <link href="blocking.css" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css?family=Muli:400,600,700" rel="prefetch">
+    <link href="https://fonts.googleapis.com/css?family=Fira+Mono&text=0123456789ABCDEFGHJKLMNPQRSTUVXY" rel="prefetch">
```
danimoh: This doesn't work.
This really only triggers the download of the specified files but they will not be applied as stylesheets.
Also `rel="prefetch"` has a low priority. Thus, a FOUC might occur.

Sorry for causing that much back and forth.
I really just meant inlining the contents of the google stylesheet into our stylesheet.
I'll create a PR towards your branch to show what I meant.

Bettelstab: Ah, you're right... well, it's quite clear to me now, but go ahead :)

Bettelstab: @danimoh I guess you're busy with faucet stuff, so I will do it tomorrow.

public/blocking.css
```diff
@@ -0,0 +1,103 @@
...
...0wLjAwMzkgMTIuNTA5NzY2LC0yMS4zMjQyMTkgYyAxLjI1NjQ3OSwtMi4xNDYyMDkgMS4yNjI3MDUsLTQuODE0ODMyIDAsLTYuOTUzMTI1IGwgLTAuMDAyLC0wLjAwMzkgLTEyLjQ4ODI4MSwtMjEuMzM3ODkwOCAtMC4wMDM5LC0wLjAwNTg2IEMgNDkuMjk3ODQzLDQuODA1MDA4NyA0Ni45OTA5NjcsMy41IDQ0LjUwNzgxMiwzLjUgWiBtIDAsNCBoIDI1LjAxNTYyNCBjIDEuMDc5ODI2LDAgMi4wNzM3OSwwLjU2OTU4MzMgMi42MTEzMjksMS40NzY1NjI1IGwgMTIuNDg2MzI4LDIxLjMzMjAzMTUgMC4wMDM5LDAuMDAzOSBjIDAuNTMyMDE2LDAuODk3NjYxIDAuNTM2MDgsMS45NzY2NTEgLTAuMDA1OSwyLjkwMjM0NCBsIC0xMi41MDE5NTQsMjEuMzA4NTk0IC0wLjAwMzksMC4wMDIgQyA0Ni41NTk2MTYsNTUuNDMwMjM2IDQ1LjU2ODYwNCw1NiA0NC40OTAyMzQsNTYgSCAxOS40OTIxODggYyAtMS4wNzk4MjYsMCAtMi4wNTc5ODYsLTAuNTYxMzkgLTIuNTc4MTI2LC0xLjQ2ODc1IGwgLTAuMDAzOSwtMC4wMDk4IC0xMi41MTU2MjQ4LC0yMS4zMzIwMzEgLTAuMDAxOTUsLTAuMDAyIGMgLTAuNTMyMDE1NywtMC44OTc2NjEgLTAuNTMyMDE1NywtMS45NzczMzkgMCwtMi44NzUgbCAwLjAwMTk1LC0wLjAwMzkgTCAxNi44ODA4NTksOC45NzY1NjI1IEMgMTcuNDE4Mzk5LDguMDY5NTgzIDE4LjQxMjM2Myw3LjUgMTkuNDkyMTg4LDcuNSBaIiBmaWxsPSIjZmZmIiBvcGFjaXR5PSIuMiIvPjxnIGNsaXAtcGF0aD0idXJsKCNoZXhDbGlwKSI+PGNpcmNsZSBpZD0iY2lyY2xlIiBjeD0iMzIiIGN5PSIzMiIgcj0iMTYiIGZpbGw9Im5vbmUiIHN0cm9rZS13aWR0aD0iMzIiIHN0cm9rZT0iIzFERTlCNiIgc3Ryb2tlLWRhc2hhcnJheT0iMTYuNjY2IDg0LjY2NiIvPjwvZz48L3N2Zz4=') !important;
```
sisou: Is `!important` at the end here really necessary?

public/blocking.css
```diff
@@ -0,0 +1,103 @@
...
+    font-family: Muli, system-ui, sans-serif;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+    text-rendering: optimizeLegibility;
+    --nimiq-gray: #F4F4F5;
+    --nimiq-gray-bg: radial-gradient(circle at bottom right, #F4F4F4, var(--nimiq-gray));
+}
+
+body {
+    color: var(--nimiq-blue);
```
sisou: `--nimiq-blue` is not defined in this stylesheet.

public/blocking.css
```diff
@@ -0,0 +1,103 @@
...
+    background-repeat: no-repeat;
+    background-position: center;
+    background-size: 100%;
+    height: 10rem;
+    width: 10rem;
+}
+
+.loading-animation+h2 {
```
sisou: Missing spaces around the `+`.

public/index.html
sisou: I'm actually fine with just making them `stylesheets` again and forgetting the inlining for now.

public/blocking.css
```diff
@@ -0,0 +1,103 @@
+html, body {
+    min-height: 100%;
```
sisou: As seen in 14757d43f3c5b1eec87cb729365242914a11b06b, not setting `height` is not enough and makes the content align to the top of the page, instead of be vertically centered.

public/blocking.css
Bettelstab: Otherwise it uses the non-working version of @nimiq/style, so yes :( But I'll check again now...

Bettelstab: Seems to work now, most probably because something you changed in the LoadingSpinner component. So I removed the !important :)

Bettelstab:
...and again @danimoh @sisou

public/index.html
danimoh: I created the PR now: https://github.com/nimiq/accounts/pull/116

PR #106, opened 2018-12-12
Enable page refresh at every step

sisou:
With these changes, and the according changes in the @nimiq/rpc and @nimiq/keyguard-client libs (already included here as package dependencies), it becomes possible for the user to refresh the page at every step throughout a request flow, without loosing the request or getting stuck.

### DIFF
--- a/client/RequestBehavior.ts
+++ b/client/RequestBehavior.ts
@@ -1,5 +1,5 @@
 import { RequestType } from '../src/lib/RequestTypes';
-import * as Rpc from '@nimiq/rpc';
+import { PostMessageRpcClient, RedirectRpcClient } from '@nimiq/rpc';
 
 export class RequestBehavior {
     public static getAllowedOrigin(endpoint: string) {
@@ -51,7 +51,7 @@ export class RedirectRequestBehavior extends RequestBehavior {
     public async request(endpoint: string, command: RequestType, args: any[]): Promise<any> {
         const origin = RequestBehavior.getAllowedOrigin(endpoint);
 
-        const client = new Rpc.RedirectRpcClient(endpoint, origin);
+        const client = new RedirectRpcClient(endpoint, origin);
         await client.init();
 
         const state: object = Object.assign({}, this._localState, { __command: command });
@@ -73,58 +73,16 @@ export class PopupRequestBehavior extends RequestBehavior {
         const origin = RequestBehavior.getAllowedOrigin(endpoint);
 
         const popup = this.createPopup(endpoint);
-        const client = new Rpc.PostMessageRpcClient(popup, origin);
+        const client = new PostMessageRpcClient(popup, origin);
         await client.init();
 
         try {
-            const result = await client.call(command, ...args);
-            client.close();
-            popup.close();
-            return result;
+            return await client.callAndPersist(command, ...args);
         } catch (e) {
-            client.close();
-            popup.close();
             throw e;
-        }
-    }
-
-    public createPopup(url: string) {
-        const popup = window.open(url, 'NimiqAccounts', this._options);
-        if (!popup) {
-            throw new Error('Failed to open popup');
-        }
-        return popup;
-    }
-}
-
-export class UrlEncodedPopupBehavior extends RequestBehavior {
-    private static DEFAULT_OPTIONS: string = '';
-    private _options: string;
-
-    constructor(options = UrlEncodedPopupBehavior.DEFAULT_OPTIONS) {
-        super(BehaviorType.POPUP);
-        this._options = options;
-    }
-
-    public async request(endpoint: string, command: RequestType, args: any[]): Promise<any> {
-        const origin = RequestBehavior.getAllowedOrigin(endpoint);
-
-        const id = Rpc.RandomUtils.generateRandomId();
-        const url = Rpc.UrlRpcEncoder.prepareRedirectInvocation(endpoint, id, '<postMessage>', command, args);
-
-        const popup = this.createPopup(url);
-        const client = new Rpc.ReceiveOnlyPostMessageRpcClient(popup, origin, id);
-        await client.init();
-
-        try {
-            const result = await client.listenFor(command);
+        } finally {
             client.close();
             popup.close();
-            return result;
-        } catch (e) {
-            client.close();
-            popup.close();
-            throw e;
         }
     }
 
@@ -141,7 +99,7 @@ export class IFrameRequestBehavior extends RequestBehavior {
     private static IFRAME_PATH_SUFFIX = '/iframe.html';
 
     private _iframe: HTMLIFrameElement | null;
-    private _client: Rpc.PostMessageRpcClient | null;
+    private _client: PostMessageRpcClient | null;
 
     constructor() {
         super(BehaviorType.IFRAME);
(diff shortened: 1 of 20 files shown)


client/RequestBehavior.ts
```diff
@@ -97,11 +97,51 @@ export class PopupRequestBehavior extends RequestBehavior {
...
+
+        try {
+            const result = await client.listenFor(command);
+            client.close();
+            popup.close();
+            return result;
+        } catch (e) {
+            client.close();
+            popup.close();
+            throw e;
```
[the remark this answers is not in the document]
sisou: Does the `finally` block still run even when the function `return`s in the `try` block?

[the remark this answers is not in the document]
sisou: I just looked it up on MDN and the answer should be yes. In that case, yes, let's do that.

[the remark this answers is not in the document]
danimoh: Yes, `finally` is pretty cool

src/lib/RpcApi.ts
```diff
@@ -62,6 +62,16 @@ export default class RpcApi {
         return client;
     }
 
+    public routerPush(routeName: string) {
```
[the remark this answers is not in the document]
sisou: They need to be accessible from every component, thus the Vue prototype property `.$rpc` lends itself perfectly to thus oppertunity, as otherwise we would have to extend the native `vue-router`, which would also be available in each component as `.$router`. These `routerPush` and `routerReplace` functions however are wrappers for the `vue-router`'s native `push` and `replace` methods. Thus it's a bit hard to come up with better names on the router instance itself, but it would still be prototype-extension, which is generally considered bad practice.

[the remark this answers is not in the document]
sisou: And it also actually has to do with the RpcApi itself, because it's used for preserving the rpc parameters of the request and responses. Thus it's not _that_ bad a place for the methods.

client/RequestBehavior.ts
```diff
@@ -77,14 +77,12 @@ export class PopupRequestBehavior extends RequestBehavior {
         await client.init();
 
         try {
-            const result = await client.call(command, ...args);
-            client.close();
-            popup.close();
-            return result;
+            return await client.callAndPersist(command, ...args);
```
sisou: Note for reviewers: this is basically where the magic is happening. The new `callAndPersist()` method makes sure that the receiving AccountsManager RpcServer writes the request into the URL after validating it, thus enabling a page refresh in the AccountsManager (after a page refresh, the request would be read from the URL as if it were a redirect request, but a special `returnUrl` parameter signals that the request should be replied to via postMessage).

src/main.ts
```diff
@@ -15,6 +15,7 @@ self.NIMIQ_IQONS_SVG_PATH = '/img/iqons.min.cc877caa.svg';
 
 const rpcApi = new RpcApi(store, staticStore, router);
 Vue.prototype.$rpc = rpcApi;
+// rpcApi ist started in App.vue->created()
```
nibhar:
```suggestion
// rpcApi is started in App.vue->created()
```

client/package.json
```diff
@@ -8,7 +8,7 @@
   "license": "Apache-2.0",
   "private": false,
   "dependencies": {
-    "@nimiq/rpc": "^0.0.11"
+    "@nimiq/rpc": "^0.1.0-beta.5"
```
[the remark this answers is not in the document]
sisou: RPC release pending here: https://github.com/nimiq/rpc/pull/8

package.json
```diff
@@ -5,16 +5,16 @@
...
         "test:unit": "vue-cli-service test:unit",
         "test": "yarn test:unit",
         "pr": "yarn lint && yarn build:ci && yarn test && cd client && yarn lint && yarn build"
     },
     "dependencies": {
-        "@nimiq/keyguard-client": "^0.1.2",
+        "@nimiq/keyguard-client": "^0.1.5",
         "@nimiq/network-client": "^0.1.0",
-        "@nimiq/rpc": "^0.0.11",
+        "@nimiq/rpc": "^0.1.0-beta.4",
```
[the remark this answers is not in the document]
sisou: RPC release pending here: https://github.com/nimiq/rpc/pull/8

src/components/Network.vue
```diff
@@ -59,7 +59,6 @@ class Network extends Vue {
                     new Nimiq.PublicKey(keyguardResult.publicKey),
                     new Nimiq.Signature(keyguardResult.signature),
                 ).serialize(),
-                keyguardRequest.networkId,
```
Bettelstab: I don't think this is related to page refresh, is it?

src/lib/RpcApi.ts
```diff
@@ -114,11 +130,25 @@ export default class RpcApi {
...
+                this.routerReplace(request);
             });
         }
     }
 
+    private _parseUrlParams(query: string) {
+        const params: {[key: string]: string} = {};
+        if (!query) return params;
+        const keyValues = query.substr(1).replace(/\+/g, ' ').split('&')
+            .map((keyValueString) => keyValueString.split('='));
```
Bettelstab: You could use URLSearchParams here.
```suggestion\
           const keyValues = new URLSearchParams(query);
```

src/views/CheckoutOverview.vue
```diff
@@ -75,7 +76,6 @@ export default class CheckoutOverview extends Vue {
             validityStartHeight,
             data: this.request.data,
             flags: this.request.flags,
-            networkId: this.request.networkId,
```
Bettelstab: unrelated

src/views/SignTransaction.vue
```diff
@@ -43,7 +43,6 @@ export default class SignTransaction extends Vue {
             validityStartHeight: this.request.validityStartHeight,
             data: this.request.data,
             flags: this.request.flags,
-            networkId: this.request.networkId,
```
Bettelstab: unrelated

src/components/Network.vue
sisou: It is, because I am updating the KeyguardClient dependency, to have the RPC features required, but the API there changed and does not include the networkId anymore. Thus, to use the new KeyguardClient, this needs to be removed (will otherwise throw a typescript error).

src/views/CheckoutOverview.vue
sisou: It is related to updating the KeyguardClient dependency, to have the RPC features required. The API there changed and does not include the networkId anymore. Thus, to use the new KeyguardClient, this needs to be removed (will otherwise throw a typescript error).

src/views/SignTransaction.vue
sisou: It is related to updating the KeyguardClient dependency, to have the RPC features required. The API there changed and does not include the networkId anymore. Thus, to use the new KeyguardClient, this needs to be removed (will otherwise throw a typescript error).

src/lib/RpcApi.ts
sisou: `URLSearchParams` decodes the params differently from what vue-router does (https://github.com/vuejs/vue-router/blob/dev/src/util/query.js#L65), which is exactly the problem I created this function for. Otherwise it would have been easy :laughing:.

Bettelstab: I see :D

PR #111, opened 2018-12-18
Checkout style adaptions

danimoh:
This PR is based on `soeren/new-checkout` and also targeting that branch.

If I find the time, I'll add more changes to this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
         "@nimiq/keyguard-client": "^0.1.2",
         "@nimiq/network-client": "^0.1.3",
         "@nimiq/rpc": "^0.0.11",
-        "@nimiq/style": "^0.4.0",
+        "@nimiq/style": "https://github.com/nimiq/nimiq-style.git#julian/hover-states",
         "@nimiq/utils": "^0.1.0",
         "@nimiq/vue-components": "https://github.com/nimiq/vue-components.git#build/accounts-new",
         "vue": "^2.5.16",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
         "@nimiq/keyguard-client": "^0.1.2",
         "@nimiq/network-client": "^0.1.3",
         "@nimiq/rpc": "^0.0.11",
-        "@nimiq/style": "https://github.com/nimiq/nimiq-style.git#julian/hover-states",
+        "@nimiq/style": "^0.4.1",
         "@nimiq/utils": "^0.1.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
         "@nimiq/keyguard-client": "^0.1.2",
         "@nimiq/network-client": "^0.1.3",
         "@nimiq/rpc": "^0.0.11",
-        "@nimiq/style": "^0.4.0",
+        "@nimiq/style": "https://github.com/nimiq/nimiq-style.git#julian/hover-states",
```
sisou: `v.0.4.1` of this package has been released now.

src/App.vue
```diff
@@ -93,7 +93,7 @@ export default class App extends Vue {
     .logo .nimiq-logo {
         height: 4rem;
         width: 4rem;
-        margin-right: 2rem;
+        margin-right: 1rem;
```
sisou: The styleguide says, the gap between the signet and the NIMIQ wordmark should be 0.382 times the signet width. So if the signet is 4rem wide, the margin should be `0.382 * 4rem = 1.528rem`.

src/views/Checkout.vue
```diff
@@ -330,7 +330,11 @@ export default class Checkout extends Vue {
         transition: opacity 300ms, z-index 300ms;
     }
 
-    .account-info.active {
+    .merchant-info-shown > :not(.account-info) {
+        filter: blur(.75rem);
```
sisou: @nibhar I think you discussed with Julian the exact blur amounts for the various elements. Can you help out here?

src/views/Checkout.vue
```diff
@@ -367,5 +371,7 @@ export default class Checkout extends Vue {
 <style>
     #logo-checkout {
         margin-left: 0.75rem;
+        text-transform: none;
+        letter-spacing: initial;
```
sisou: These rules probably don't hurt, but why are they necessary?

src/App.vue
danimoh: Yes, you're right. Actually the logo text fixes were a quick fix to roughly match an SVG logo Julian sent me.
The logo label does now perfectly match that svg.

src/views/Checkout.vue
danimoh: `text-transform` and `letter-spacing` are set on the parent with class `logo` to non-initial values.
I moved this to `App.vue` as a general style rule now.

PR #116, opened 2018-12-29
Integrate fonts css into blocking.css

danimoh:
Copied styles from `https://fonts.googleapis.com/css?family=Muli:400,600,700` and `https://fonts.googleapis.com/css?family=Fira+Mono&text=0123456789ABCDEFGHJKLMNPQRSTUVXY` but commented Vietnamese out.
Also manually added `woff` fallback :dog:

### DIFF
--- a/public/blocking.css
+++ b/public/blocking.css
@@ -106,20 +106,18 @@ body {
 
 /*
  * Style copied from https://fonts.googleapis.com/css?family=Muli:400,600,700
- * But currently commented-out vietnamese characters (this might be not necessary due to specified unicode-range though)
- * Also manually added woff fallback files.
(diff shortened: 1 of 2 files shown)


danimoh:
The links might change.
In this case, we'd have to change the links also.
But I wouldn't expect this to happen too often and if it does, we can still go back to linking to the css files.

Bettelstab:
What about vietnamese characters used as account labels?

danimoh:
Good point!
I enabled the Vietnamese font now.
From my understanding of `unicode-range`, it _should_ also only be downloaded if Vietnamese characters are used in the page.
Thus, this should not have an impact on loading time in the general case.

PR #126, opened 2019-02-06
HTTPS

svub:
Note on how to enable HTTPS in production.
`package.json` update enabling HTTPS in dev mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "version": "0.1.0",
     "private": true,
     "scripts": {
-        "serve": "vue-cli-service serve",
+        "serve": "vue-cli-service serve --https",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
     "version": "0.1.0",
     "private": true,
     "scripts": {
-        "serve": "vue-cli-service serve",
+        "serve": "vue-cli-service serve --https",
```
sisou: It's not the AccountsManager that needs to be served through HTTPS locally (if it would, however, the Keyguard and app would also have to, locally), but the developer's own app that he/she develops locally, but is testing with https://accounts.nimiq-testnet.com (also only for redirects, as @nibhar and I think).

PR #129, opened 2019-02-12
Fix automatic re-starting requests when returning from Keyguard

sisou:


sisou:
Added the `Export.vue` in the force-push.

PR #132, opened 2019-02-12
Use new Loader for success screens

sisou:


src/views/SignTransactionSuccess.vue
```diff
@@ -1,12 +1,12 @@
...
-            <Success
-                text="Your transaction[br]is ready!"
-                buttonText="Send now"
-                :disabled="!isTxPrepared"
-                @continue="done"
-            />
+            <Loader state="success">
+                <template slot="success">
+                    <div class="success nq-icon"></div>
+                    <h1 class="title nq-h1">Sending your<br>transaction now.</h1>
```
Bettelstab: I'm not quite happy with stating it is sent now, while it isn't before returning to Safe.

sisou: Especially with the new pico consensus, we could just as well send it here, just like during checkout.

sisou:
Rebased on `master` to fix merge conflicts.

Bettelstab:
Please rename Loader to e.g. StatusIndicator :)

sisou:
#135 also changes the Loader component quite a bit. I'd rather wait with the renaming until after both are merged. I created #142 to not forget about it.

PR #135, opened 2019-02-13
New onboarding / signup

sisou:


src/views/Rename.vue
```diff
@@ -133,7 +133,7 @@ export default class Rename extends Vue {
 
     @Emit()
     private close() {
-        this.$rpc.reject(new Error('CANCEL'));
+        this.$rpc.reject(new Error('CANCELED'));
```
Bettelstab: This should also be a constant.

src/views/SignMessage.vue
```diff
@@ -100,7 +100,7 @@ export default class SignMessage extends Vue {
 
     @Emit()
     private close() {
-        this.$rpc.reject(new Error('CANCEL'));
+        this.$rpc.reject(new Error('CANCELED'));
```
Bettelstab: constant

src/views/Checkout.vue
```diff
@@ -197,23 +199,16 @@ export default class Checkout extends Vue {
...
-        const client = this.$rpc.createKeyguardClient();
-        client.import(request).catch(console.error); // TODO: proper error handling
+        if (useReplace) this.$rpc.routerReplace(RequestType.ONBOARD);
+        else this.$rpc.routerPush(RequestType.ONBOARD);
     }
 
     @Emit()
     private close() {
-        this.$rpc.reject(new Error('CANCEL'));
+        this.$rpc.reject(new Error('CANCELED'));
```
Bettelstab: Constant :)

src/lib/RpcApi.ts
```diff
@@ -76,7 +77,7 @@ export default class RpcApi {
     }
 
     public reject(error: Error) {
-        const ignoredErrors = [ 'CANCEL', 'Request aborted' ];
+        const ignoredErrors = [ 'CANCELED', 'Request aborted' ];
```
Bettelstab: Constants

src/lib/RpcApi.ts
```diff
@@ -191,7 +192,7 @@ export default class RpcApi {
                 // Recover state
                 this._recoverState(state);
 
-                if (error.message === 'CANCEL') {
+                if (error.message === 'CANCELED') {
```
Bettelstab: Constant

src/views/SignupSuccess.vue
```diff
@@ -1,170 +1,93 @@
...
+    ACCOUNT_DEFAULT_LABEL_LEDGER,
+} from '@/lib/Constants';
+
+@Component({components: {SmallPage}})
 export default class SignupSuccess extends Vue {
-    private static readonly STEPS_KEYGUARD_SIGNUP = 6;
-    private static readonly STEPS_LEDGER_SIGNUP = 3;
-
     @Prop({ default: null })
-    public createResult!: CreateResult;
```
Bettelstab: actually I would prefer to remove ambient types completely

Bettelstab: PS: TO make importing less cumbersome, we could export/import KeyguardRequest (equals same as before) instead of all the requests and results for themselves.

sisou: Removing ambient types should be it's own PR.

PR #140, opened 2019-02-15
Choose address request

Bettelstab:
Working. Currently, `processWallets()` is a copy from Checkout. Maybe make both use a mixin with that functionality.


sisou:
In the SignMessage PR, the `processedWallets` is moved into the store as a getter.

PR #141, opened 2019-02-16
Use @nimiq/core-web and get rid of ambient types

Bettelstab:
Disclaimer: Broken test architecture.
Needs https://github.com/nimiq/network/pull/4 and https://github.com/nimiq/keyguard-next/pull/202 to be merged.


package.json
```diff
@@ -13,13 +13,14 @@
         "postinstall": "cd client && yarn"
     },
     "dependencies": {
-        "@nimiq/keyguard-client": "0.1.6",
-        "@nimiq/network-client": "^0.1.3",
+        "@nimiq/network-client": "^0.1.5-beta2",
```
sisou: Downgrade of network-client? Is this correct?

src/App.vue
```diff
@@ -16,6 +16,7 @@
 <script lang="ts">
 import { Component, Watch, Vue } from 'vue-property-decorator';
 import { State } from 'vuex-class';
+import * as KClient from '@nimiq/keyguard-client';
```
sisou: This import is not used, AFAIK, so can be removed.

src/lib/CookieJar.ts
```diff
@@ -1,5 +1,5 @@
 // tslint:disable no-bitwise no-shadowed-variable
-
+import Nimiq from '@nimiq/core-web';
```
sisou: I liked that empty line.

src/lib/WalletInfoCollector.ts
```diff
@@ -128,7 +129,7 @@ export default class WalletInfoCollector {
             WalletInfoCollector._keyguardClient = WalletInfoCollector._keyguardClient || new KeyguardClient();
         } else if (walletType === WalletType.LEDGER) {
             WalletInfoCollector._wasmInitializationPromise =
-                WalletInfoCollector._wasmInitializationPromise || Nimiq.WasmHelper.doImportBrowser();
+                WalletInfoCollector._wasmInitializationPromise || Nimiq.WasmHelper.doImport();
```
sisou: @danimoh I believe with your changes in core in 1.4.3 this is now simpler to track, as `doImport()` always returns the same promise, right? So `_wasmInitializationPromise` is not needed anymore.

package.json
Bettelstab: How is that a downgrade?

sisou: Ah, I was confused, because you changed the order of the dependencies. Please fix the order. keyguard-client comes before network-client in the alphabet.

Bettelstab:
Bundle sizes are much larger than thought, so we shouldn't merge this.
The changes regarding KeyguardRequest namespace => imports have to go in an own PR meanwhile.

PR #144, opened 2019-02-18
Network and keyguard client updates

Bettelstab:
Includes some changes due to https://github.com/nimiq/keyguard-next/pull/200, which needs to be merged first.


src/lib/RpcApi.ts
```diff
@@ -182,15 +182,14 @@ export default class RpcApi {
                 this._recoverState(state);
 
                 // Set result
-                result.kind = command;
```
sisou: I did this because it was a type error in TS, but I'm not actually sure if this doesn't break something somewhere. I looked, but couldn't find code that uses it.

Bettelstab: Should be safe to remove it for now.

PR #146, opened 2019-02-22
Sign Message

sisou:


src/views/SignMessage.vue
```diff
@@ -12,97 +26,152 @@
...
-    protected sendKeyguardRequest(walletInfo: WalletInfo, accountInfo: AccountInfo) {
-        const request: KSignMessageRequest = {
+    private async accountSelected(walletId: string, address: string) {
+        const walletInfo = this.findWallet(walletId);
+        if (!walletInfo) {
+            // XXX Should we really return an error here and when checking the address below,
+            // or would it enable malicous sites to query for stored walletIds and addresses?
+            // We could also quietly ignore any unavailable pre-set walletId and address and give
+            // the user the option to chose as if it was not pre-set.
+            this.$rpc.reject(new Error('WalletId not found'));
```
Bettelstab: Right, better no error. I like your suggestion.

src/views/SignMessage.vue
```diff
@@ -12,97 +26,152 @@
...
+            return;
+        }
+
+        this.showAccountSelector = true;
+
+    }
 
-    protected sendKeyguardRequest(walletInfo: WalletInfo, accountInfo: AccountInfo) {
-        const request: KSignMessageRequest = {
+    private async accountSelected(walletId: string, address: string) {
```
Bettelstab: Maybe `selectAccount` is a better name.

src/views/SignMessage.vue
```diff
@@ -12,97 +26,152 @@
...
         return Utf8Tools.stringToUtf8ByteArray(this.request.message);
     }
 
-    private async created() {
-        // Only run when not returning from Keyguard
-        if (this.keyguardResult) return;
+    private goToOnboarding(useReplace?: boolean) {
+        // Redirect to onboarding
+        staticStore.originalRouteName = RequestType.SIGN_MESSAGE;
+        if (useReplace) this.$rpc.routerReplace(RequestType.ONBOARD);
```
Bettelstab: Didn't we agree to not use one-line ifs if it's not a return? :)

src/views/SignMessage.vue
sisou: It's literally the function that get assigned as the handler for the `account-selected` event of the account selector. I'd rather name it `onAccountSelected` then. Is that what you mean?

Bettelstab: That doesn't really fit, because it's also called in `created()` if the request contains the signer.

src/views/SignMessage.vue
```diff
@@ -52,32 +52,33 @@ export default class SignMessage extends Vue {
...
-            // or would it enable malicous sites to query for stored walletIds and addresses?
-            // We could also quietly ignore any unavailable pre-set walletId and address and give
+            // We can also return an error here and when checking the address below,
+            // but it would enable malicous sites to query for stored walletIds and addresses.
+            // Instead we quietly ignore any unavailable pre-set walletId and address and give
             // the user the option to chose as if it was not pre-set.
-            this.$rpc.reject(new Error('WalletId not found'));
+            this.showAccountSelector = true;
+            if (!isFromRequest) throw new Error(`UNEXPECTED: Selected walletId was not found: ${walletId}`);
             return;
```
Bettelstab: Hm.. I think this error (and the one below) are *so* unexpected, it doesn't make sense to even handle those cases.

Issue #112, opened 2018-12-19
Disable ‘back to safe’ in onboarding.


Bettelstab:
Disable ‘back to safe’ if there is no account. 
How: probably as request parameter. Or always?

PR #149, opened 2019-02-26
Make back button in onboarding optional

Bettelstab:
Solves #112: Going back to Safe is useless as Safe will redirect again to Onboarding.
Will need another PR in Safe which uses this option.


src/lib/RequestTypes.ts
```diff
@@ -328,10 +338,11 @@ export class AccountsRequest {
                     ),
                 } as ParsedCheckoutRequest;
             case RequestType.ONBOARD:
-                request = request as SimpleRequest;
+                request = request as OnboardingRequest;
                 return {
                     kind: RequestType.ONBOARD,
                     appName: request.appName,
+                    hideBack: request.hideBack === true,
```
sisou: Another possibility (not saying it's better, just possible :wink:)
```suggestion
                    hideBack: !!request.hideBack,
```

src/lib/RequestTypes.ts
```diff
@@ -19,14 +19,20 @@ export enum RequestType {
...
 }
 
 export interface SimpleResult {
     success: true;
 }
 
+export interface OnboardingRequest {
+    kind?: RequestType.ONBOARD;
+    appName: string;
+    hideBack?: boolean;
```
sisou: I propose the name `disableBack` for the parameter.

src/lib/RequestTypes.ts
Bettelstab: Never liked that. No no no :D

Bettelstab: Especially, in this case I want that value to be exactly true, and not "nipir32pnp132r" or whatever would be possible with `!!` ;)

PR #150, opened 2019-02-26
set secure and samesite strict in cookie

sisou:


sisou:
I guess this breaks testing in non-HTTPS localhost and test environments.

Bettelstab:
Yes. I did some research there how it might possible to get it running, but end up suggesting to remove the broken tests, as the remaining ones are not as fine-grained, but still one would notice a malfunction.

danimoh:
Instead of removing the test, `fill` might only add the `Secure` part if not served from localhost

PR #154, opened 2019-03-08
Add success title slide-up to Loader

sisou:


src/components/Loader.vue
```diff
@@ -395,4 +395,23 @@ export default Loader;
         max-width: 70%;
         opacity: 1;
     }
+
+    .loader.exit-transition {
+        /* animation: exit-transition 600ms 1s; */
```
Bettelstab: delete this...

src/components/Loader.vue
```diff
@@ -395,4 +395,23 @@ export default Loader;
...
+    }
+
+    @keyframes exit-transition {
+        from { transform: scale(1); opacity: 1; }
+        80%  { opacity: 0; }
+        to   { transform: scale(0); opacity: 0; }
+    }
```
Bettelstab: ...until here, I guess?

src/components/Loader.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="loader" :class="showLoadingBackground && (lightBlue ? 'nq-light-blue-bg' : 'nq-blue-bg')">
+    <div class="loader" :class="[showLoadingBackground && (lightBlue ? 'nq-light-blue-bg' : 'nq-blue-bg'), state === 'success' && 'exit-transition']">
```
danimoh: Would be nicer to use
```suggestion
    <div class="loader" :class="{ 'nq-blue-bq': showLoadingBackground && !lightBlue, 'nq-light-blue-bg': showLoadingBackground && lightBlue, 'exit-transition': state === 'success' }">
```

src/components/Loader.vue
```diff
@@ -395,4 +395,23 @@ export default Loader;
...
+        /* animation: exit-transition 600ms 1s; */
+    }
+
+    @keyframes exit-transition {
+        from { transform: scale(1); opacity: 1; }
+        80%  { opacity: 0; }
+        to   { transform: scale(0); opacity: 0; }
+    }
+
+    .loader.exit-transition .success .icon-row {
```
danimoh: What's actually the reason for the unscoped second `<style>` tag and manual scoping instead?

src/components/Loader.vue
danimoh: The whole `exit-transition` class + logic can then also be removed, as currently it only affects `.success .icon-row` which is only visible on success anyways.

src/components/Loader.vue
sisou: Slots do not inherit scoped styles. Thus when we use custom HTML for slots from outside, those can only be styled with global styles.

PR #159, opened 2019-03-11
Integrate faucet in testnet version of checkout

Bettelstab:
Includes performance improvement, better typing and error handling.
Before merging, I would like to try it with real PicoConsensus first to see if we can use `subscribe()` now instead of relying on timeouts.


src/views/Checkout.vue
```diff
@@ -147,6 +160,73 @@ export default class Checkout extends Vue {
...
+                console.log('check balance');
+                if (this.firstAddress.balance && newBalance > this.firstAddress.balance) {
+                    this.firstAddress.balance = newBalance; // still referenced in wallet's accounts map
+                    const updatedWallet = this.wallets[0];
+                    // Update IndexedDB
+                    await WalletStore.Instance.put(updatedWallet);
+                    // Update Vuex
+                    this.$addWallet(updatedWallet);
+                    this.isTappingFaucet = false;
+                }
```
danimoh: Why the duplicate balance fetch via `this.getBalances(true)` and `network.getBalances`?
Also `this.getBalances` invokes `network.connectPico` which disconnects the old network client.
Might be that this could interfere with the `network.getBalances` call?

I'd use only `this.getBalance` (and also await it) which removes the double fetch and necessity to update the store here.
The check whether the balance changes should then happen against a stored old value.

src/views/Checkout.vue
```diff
@@ -147,6 +160,73 @@ export default class Checkout extends Vue {
...
+                const newBalance = [ ...balances.values() ][0];
+                console.log('check balance');
+                if (this.firstAddress.balance && newBalance > this.firstAddress.balance) {
+                    this.firstAddress.balance = newBalance; // still referenced in wallet's accounts map
+                    const updatedWallet = this.wallets[0];
+                    // Update IndexedDB
+                    await WalletStore.Instance.put(updatedWallet);
+                    // Update Vuex
+                    this.$addWallet(updatedWallet);
+                    this.isTappingFaucet = false;
```
danimoh: Missing a return statement. Currently it's an endless loop.

src/views/Checkout.vue
```diff
@@ -147,6 +160,73 @@ export default class Checkout extends Vue {
...
+                }
+                setTimeout(checkForArrivedBalance, 1000);
```
danimoh: As this invokes `network.connectPico` which reconnects the connection, doing this every second could be quite hardcore

src/lib/Faucet.ts
```diff
@@ -0,0 +1,44 @@
...
+                'address': recipientAddress,
+                'g-recaptcha-response': captchaToken,
+            }),
+        }).then((finalResponse) => finalResponse.json());
+        if (!response.success) {
+            // FIXME
+            const error = new Error('Faucet error') as any;
+            error.error = response.error;
+            error.msg = response.msg;
+            throw error;
```
danimoh: As you're not using the error codes, these lines can be reduced to:
```js
throw new Error(response.msg || 'Failed to claim funds');
```

src/views/Checkout.vue
```diff
@@ -147,6 +160,73 @@ export default class Checkout extends Vue {
...
+            };
+            setTimeout(checkForArrivedBalance, 1000);
+        } catch (e) {
+            const errorMessage = e.msg // faucet specific error message
+                || e.message // message of Error constructor
+                || 'Failed to claim funds.';
+            this.faucetError = errorMessage;
```
danimoh: Can be reduced to 
```js
this.faucetError = e.message;
```
if changing the error throwing in the Faucet as suggested above.

src/views/Checkout.vue
```diff
@@ -230,6 +310,14 @@ export default class Checkout extends Vue {
         });
     }
 
+    private get firstAddress(): AccountInfo {
+        if (!this._firstAddress) {
+            this._firstAddress = this._firstAddress || [ ...this.wallets[0].accounts.values() ][0];
```
danimoh: The `this._firstAddress || ` part is unnecessary.

danimoh:
Closing this, as the nimiq safe now provides a faucet

PR #166, opened 2019-03-15
Config enhancement and refactoring

Bettelstab:
* Add and use config parameter for network endpoint
* Move configuration of `nimiqOrigins` (now `privilegedOrigins`) to config files and delete old `/lib/Config.ts`


config.mainnet.ts
```diff
@@ -3,4 +3,6 @@ import { NETWORK_MAIN } from './src/lib/Constants';
 export default {
     keyguardEndpoint: 'https://keyguard.nimiq.com',
     network: NETWORK_MAIN,
+    networkEndpoint: 'https://network.nimiq.com',
+    privilegedOrigins: [ 'safe.nimiq.com' ],
```
sisou: An origin includes the protocol, so I think this should be `https://safe.nimiq.com`.

config.testnet.ts
```diff
@@ -3,4 +3,6 @@ import { NETWORK_TEST } from './src/lib/Constants';
 export default {
     keyguardEndpoint: 'https://keyguard.nimiq-testnet.com',
     network: NETWORK_TEST,
+    networkEndpoint: 'https://network.nimiq-testnet.com',
+    privilegedOrigins: [ 'safe.nimiq-testnet.com' ],
```
sisou: Same here.

jest.config.js
```diff
@@ -12,7 +12,8 @@ module.exports = {
     '^.+\\.tsx?$': 'ts-jest'
   },
   moduleNameMapper: {
-    '^@/(.*)$': '<rootDir>/src/$1'
+    '^@/(.*)$': '<rootDir>/src/$1',
+    "config": "<rootDir>/config.local.ts",
```
sisou: We usually use single quotes.

sisou:
In your commit https://github.com/nimiq/network/commit/0a72bc73882b2badc3ef4bd523eaa94c62bce14b in the NetworkClient from 2nd Feb you removed the `-next` from `network-next` of the default domain for the network.

I don't remember what the changes are between `network` and `network-next` deployments, but there _were_ some incompatibilities, so we need to make sure all apps that use the network iframe work with the `next` version. 

I think the `getsome` app uses the iframe, as might other apps. So before we deploy the `network` again, we need to fix the other apps, too (Or stay with `network-next` for now.)

(In the mainnet, even more apps might be reliant on the old network API, e.g. the Ledger Promo, too?

Bettelstab:
Are you sure there are breaking changes?
At the moment, there is no network-next anymore (it's forwarding transparently), so we would have to ask Curd to reactivate it.

sisou:
Are you sure? I thought it still existed as it's own deployment?

Bettelstab:
No, since Tico Blockchain that should be the case. So testnet network is the new version (formerly `-next`) and mainnet the old one.
Btw, `getsome` is broken anyways when we deploy to mainnet. `Promo` too. To Fix 1-2 changes in network API should be the least amount of work ;)

Issue #168, opened 2019-03-15
Request type update


nibhar:
- [ ] remove `kind` from requests.
- [ ] provide Basic and Simple Request types and inherit from them.
- [ ] move request parsing to its own file.

sisou:
Resolved by https://github.com/nimiq/accounts/pull/169.

PR #169, opened 2019-03-18
Type Update

nibhar:
this PR updates the request types used.

- types are separated into public and non-public types with their own files each
- the AccountsRequest class has its own file
- public requests no longer have the `kind` field as it was redundant.
- adds BasicRequest, ParsedBasicRequest, SimpleRequest and ParsedSimpleRequest interfaces and inherits from them

resolves all points of #168


sisou:
What do you think about renaming the `AccountsRequest` class to `RequestParser`, just like in the keyguard? I think the name AccountsRequest does not tell what the class does.

nibhar:
I wanted to keep the changes to a minimum, but I would agree.

src/lib/AccountsRequest.ts
```diff
@@ -0,0 +1,180 @@
...
+    ParsedSimpleRequest,
+    ParsedSignTransactionRequest,
+    ParsedCheckoutRequest,
+    ParsedSignMessageRequest,
+    ParsedRenameRequest,
+    ParsedRpcRequest,
+} from './RequestTypes';
+
+export class AccountsRequest {
+    public static parse(request: RpcRequest, state: State, requestType?: RequestType): ParsedRpcRequest | null {
```
sisou: The `requestType` parameter should no longer be optional.

src/lib/RpcApi.ts
```diff
@@ -3,7 +3,9 @@ import { BrowserDetection } from '@nimiq/utils';
 import { RootState } from '@/store';
 import { Store } from 'vuex';
 import Router from 'vue-router';
-import { AccountsRequest, RequestType, RpcRequest, RpcResult } from '@/lib/RequestTypes';
+import { RequestType } from './RequestTypes';
+import { AccountsRequest } from './AccountsRequest';
+import { RpcRequest, RpcResult } from './PublicRequestTypes';
```
sisou: I _really_ dont like that the request types are now split between two files, public and internal. Is that really necessary? What benefits does it bring?

Bettelstab: It clearly separates the types which are relevant for users of the client from the ones which are just internally used.

PR #178, opened 2019-03-24
Add cookie monster service worker

Bettelstab:
Adds a service worker which intercepts all network requests to the same host as its own and prevents cookie data from being transmitted.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -29,6 +29,15 @@ export default class App extends Vue {
     private isRequestLoaded = false;
 
     public async created() {
+        // Register service worker if necessary (and possible).
+        if ('serviceWorker' in navigator) {
+            navigator.serviceWorker.register('./ServiceWorker.js', {
+                scope: './',
+            }).then((reg) => {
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -29,6 +29,15 @@ export default class App extends Vue {
     private isRequestLoaded = false;
 
     public async created() {
+        // Register service worker if necessary (and possible).
+        if ('serviceWorker' in navigator) {
+            navigator.serviceWorker.register('./ServiceWorker.js', {
+                scope: './',
```
sisou: Keep in mind, that the `App.vue` component _always_ runs, also when the entry point is _not_ the root URL. All entry points that _we_ create are calling the root URL _currently_, but we cannot say that this will be the case for all user-created and future cases.

Is it thus possible, to set the scope to `'/'` instead?

Bettelstab: Should be possible, as long as the ServiceWorker is located in root, yes.

PR #179, opened 2019-03-24
Use requestTxReceipts in WalletInfoCollector

### LINKED ISSUE #94: Detect active accounts by transaction history (and balance)
Instead of just by balance.

Bettelstab:
Needs https://github.com/nimiq/nano-api/pull/6 and https://github.com/nimiq/network/pull/6 to be merged first.

Resolves #94.


src/lib/WalletInfoCollector.ts
```diff
@@ -92,14 +92,16 @@ export default class WalletInfoCollector {
...
+            // find accounts with a balance > 0 and/or transaction receipts
             foundAccounts = [];
             const balances = await WalletInfoCollector._getBalances(derivedAccounts!);
             for (const account of derivedAccounts!) {
                 const balance = balances.get(account.address);
-                if (balance === undefined || balance === 0) continue;
-                foundAccounts.push(account);
+                const hasBalance = balance !== undefined && balance !== 0;
+                const receipts = await NetworkClient.Instance.requestTransactionReceipts(account.address);
+                if (hasBalance || receipts.length > 0) {
```
sisou: Please only get receipts when `hasBalance` is false, as the receipt getting creates unnecessary network congestion and delay otherwise.

Another problem we will encounter is rate-limiting for receipts getting. `FullConsensusAgent` has a rate-limit of 30/min for transaction-receipts requests: https://github.com/nimiq-network/core/blob/master/src/main/generic/consensus/full/FullConsensusAgent.js#L469, https://github.com/nimiq-network/core/blob/master/src/main/generic/consensus/full/FullConsensusAgent.js#L591.

Thus the question is if we can switch the peer that we request from, if we have another full peer available.

This brings me to the third problem: Receipts getting only works, when we are actually connected to a full node (which is not always the case, as found in the Safe sometimes), so this needs error handling.

Bettelstab: Puh. Sounds like stuff that should be handled elsewhere...

src/lib/WalletInfoCollector.ts
```diff
@@ -92,14 +92,19 @@ export default class WalletInfoCollector {
...
             const balances = await WalletInfoCollector._getBalances(derivedAccounts!);
             for (const account of derivedAccounts!) {
                 const balance = balances.get(account.address);
-                if (balance === undefined || balance === 0) continue;
-                foundAccounts.push(account);
+                if (balance !== undefined && balance !== 0) {
+                    foundAccounts.push(account);
+                    continue;
+                }
+                const receipts = await NetworkClient.Instance.requestTransactionReceipts(account.address);
```
danimoh: Instead of `await` in the for loop, use `Promise.all` by executing the loop code as promises for each derived account

test.ts
```diff
@@ -0,0 +1,10 @@
+interface Foo {
+    mounted?(): void;
+}
+
+
+class MyFoo implements Foo {
+    private mounted() {
+        return 123;
+    }
+}
```
danimoh: :sweat_smile:

Bettelstab:
Note: Before merging this, we should redeploy network.

src/views/LoginSuccess.vue
```diff
@@ -81,8 +90,14 @@ export default class LoginSuccess extends Vue {
             })),
         };
 
-        this.title = 'Your account is ready.';
-        this.state = Loader.State.SUCCESS;
+        if (this.gotError) {
+            this.state = Loader.State.WARNING;
+            this.title = `Your addresses may be
+incomplete.`;
```
sisou: I think it's bad UX when the user gets automatically redirected after seeing a warning. I think in case of the warning, the user needs to explicitly click a "Ok, check my account" action button in the warning screen and _not_ get automatically redirected back to the Safe.

sisou: Additionally, you should set the `status` on the `Loader` to add some more text, for example: "We might have missed used addresses that have no balance."

sisou: But only the above text, if that was actually the error coming from the WalletInfoCollector. If another core error happened, that should be displayed.

src/views/LoginSuccess.vue
```diff
@@ -66,6 +69,12 @@ export default class LoginSuccess extends Vue {
         });
     }
 
+    private _onCollectAddressesError(error: Error) {
+        this.gotError = true;
+        // FIXME this.$raven is undefined
+        // this.$raven.captureException(error);
```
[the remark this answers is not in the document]
Bettelstab: No :(

danimoh:
Note, that I'll add a warning in `SignupLedger` once this PR is otherwise ready to merge

src/lib/WalletInfoCollector.ts
```diff
@@ -27,6 +27,8 @@ export default class WalletInfoCollector {
         initialAccounts?: BasicAccountInfo[],
         // tslint:disable-next-line:no-empty
         onUpdate: (walletInfo: WalletInfo, currentlyCheckedAccounts: BasicAccountInfo[]) => void = () => {},
+        // tslint:disable-next-line:no-empty
+        onError: (error: Error) => void = () => {},
```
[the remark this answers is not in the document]
Bettelstab: I'll better change the result type of collectWalletInfo then, to contain such an error. But for errors not specific to single receipt requests the standard procedure should be sufficient: try and catch, like it's already done in LoginSucces.

src/lib/WalletInfoCollector.ts
```diff
@@ -92,20 +94,39 @@ export default class WalletInfoCollector {
...
-                }
-                const receipts = await NetworkClient.Instance.requestTransactionReceipts(account.address);
-                if (receipts.length > 0) {
-                    foundAccounts.push(account);
+                    foundAddresses.push(account.address);
                 }
             }
+
+            // for accounts with balance 0 check if there are transactions
+            const addressesToCheck = derivedAccounts.map((account) => account.address)
```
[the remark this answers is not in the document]
sisou: That's unreasonable. There is no measurable difference between accessing an object property and an array directly. Since there is no computation involved in getting the address here, I think adding yet another variable is unnecessary.

src/views/LoginSuccess.vue
```diff
@@ -90,20 +94,31 @@ export default class LoginSuccess extends Vue {
...
     private get status() {
         return !this.retrievalFailed ? 'Connecting to Nimiq...' : 'Address retrieval failed. Retrying...';
     }
+
+    private get message() {
+        return this.receiptsError && 'We might have missed used addresses that have no balance.';
+    }
+
+    private get action() {
+        return this.receiptsError && 'Check account';
```
sisou: I'd call this action `Go to account` I think, because with 'Check account' I would expect that the app does another check.

sisou:
Updated network (with @nimiq/nano-api v0.2.1) is deployed.

Bettelstab:
@danimoh Otherwise ready to merge, I guess. (After my next commit)

src/lib/WalletInfoCollector.ts
```diff
@@ -128,7 +161,8 @@ export default class WalletInfoCollector {
 
     private static _initializeDependencies(walletType: WalletType): void {
         WalletInfoCollector._networkInitializationPromise =
-            WalletInfoCollector._networkInitializationPromise || NetworkClient.Instance.init();
+            WalletInfoCollector._networkInitializationPromise
+                || (NetworkClient.createInstance(Config.networkEndpoint)).init();
```
sisou:
```suggestion
                || NetworkClient.createInstance(Config.networkEndpoint).init();
```

src/lib/WalletInfoCollector.ts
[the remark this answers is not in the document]
danimoh: Of course there is no performance difference, it's a question of code style.

PR #180, opened 2019-03-25
Omit default labels in cookie

sisou:
Testing of the removal and re-setting of the default labels is done via the byte-check of the encoded cookie (the labels are empty in the cookie bytes).

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -14,6 +14,7 @@ module.exports = {
   moduleNameMapper: {
     '^@/(.*)$': '<rootDir>/src/$1',
     'config': '<rootDir>/config.local.ts',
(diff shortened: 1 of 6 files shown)


jest.config.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
   moduleNameMapper: {
     '^@/(.*)$': '<rootDir>/src/$1',
     'config': '<rootDir>/config.local.ts',
+    '@nimiq/iqons/dist/iqons-name.min.js': '@nimiq/iqons/dist/iqons-name.cjs.js',
```
Bettelstab: ha :D

sisou: Jup, simple as that :D

PR #183, opened 2019-03-30
Add add-address flow for ledger

danimoh:
Initilly I tried to refactor `SignupLedger` such that `AddAddresLedger` and `AddAddressLedger` would inherit from a common abstract base class. But in the end it turned out more beautiful to separate bothm implementations.

Note that I still want to improve the visual style of this flow (e.g. state transitions, loading spinner) but will do this together with style changes to `SignupLedger`

### DIFF
--- a/src/views/AddAddressLedger.vue
+++ b/src/views/AddAddressLedger.vue
@@ -16,7 +16,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Emit, Vue } from 'vue-property-decorator';
+import { Component, Vue } from 'vue-property-decorator';
 import { PageBody, SmallPage } from '@nimiq/vue-components';
 import LedgerUi from '../components/LedgerUi.vue';
 import Loader from '../components/Loader.vue';
@@ -95,10 +95,8 @@ export default class AddAddressLedger extends Vue {
             label: selectedAccount.label,
         };
         this.$rpc.resolve(result);
-        this.$emit('done');
     }
 
-    @Emit()
     private close() {
         this.$rpc.reject(new Error(ERROR_CANCELED));
     }
--- a/src/views/AddAddressLedger.vue
+++ b/src/views/AddAddressLedger.vue
@@ -1,0 +1,112 @@
+<template>
+    <div class="container">
+        <SmallPage>
+            <LedgerUi v-if="state === 'ledger-interaction'"></LedgerUi>
+            <IdenticonSelector v-else-if="state === 'identicon-selection'" :accounts="addressesToSelectFrom"
+                               :confirmAccountSelection="false" @identicon-selected="_onAddressSelected">
+            </IdenticonSelector>
+            <Loader v-if="state === 'finished'" state="success" title="Address Added"></Loader>
+        </SmallPage>
+
+        <button class="global-close nq-button-s" @click="close">
+            <span class="nq-icon arrow-left"></span>
+            Back to {{request.appName}}
+        </button>
+    </div>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Vue } from 'vue-property-decorator';
+import { PageBody, SmallPage } from '@nimiq/vue-components';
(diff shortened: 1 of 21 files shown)


src/views/SignupLedger.vue
```diff
@@ -69,7 +68,7 @@ export default class SignupLedger extends Vue {
 
     private get loaderTitle() {
         return this.state !== SignupLedger.State.FINISHED
-            ? 'Fetching Accounts'
+            ? 'Fetching Account'
```
sisou: I guess this is supposed to be `Fetching Addresses`?

src/views/AddAddressLedger.vue
```diff
@@ -0,0 +1,112 @@
...
+        let startIndex = 0;
+        const newestAddress = Array.from(account.accounts.values()).pop();
+        if (newestAddress) {
+            const newestIndex = LedgerApi.getKeyIdForBip32Path(newestAddress.path);
+            startIndex = (newestIndex !== null ? newestIndex : -1) + 1;
+        }
+
+        const pathsToDerive = [];
+        for (let keyId = startIndex; keyId < startIndex + ACCOUNT_MAX_ALLOWED_ADDRESS_GAP; ++keyId) {
+            pathsToDerive.push(LedgerApi.getBip32PathForKeyId(keyId));
```
sisou: I know this is not a problem with this PR specifically, but with the LedgerUi class: I think the last number in the derivation path should not be called `keyId`, as we already use that for the _ID_ of the key/wallet/account in the store. What you mean here is an _account index_ or _addresss index_. Emphasis on _index_, not _id_.

src/views/AddAddressLedger.vue
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        const derivedAddressInfos = await LedgerApi.deriveAccounts(pathsToDerive, this.request.walletId);
```
sisou: Oh, the new implementations that already use `account/address` instead of `wallet/account` will fuck us so bad when we want to bulk-refactor the code base... :(

But that is a problem in all recently added code...

src/views/AddAddressLedger.vue
```diff
@@ -0,0 +1,112 @@
...
+        await WalletStore.Instance.put(this.account);
+
+        this.state = AddAddressLedger.State.FINISHED;
+        await new Promise((resolve) => setTimeout(resolve, Loader.SUCCESS_REDIRECT_DELAY));
+        const result: Address = {
+            address: userFriendlyAddress,
+            label: selectedAccount.label,
+        };
+        this.$rpc.resolve(result);
+        this.$emit('done');
```
sisou: What is the use-case for this event, after `$rpc.resolve()` is already called?

src/views/SignupLedger.vue
danimoh: Do you like that more?
It's either fetching the whole account or fetching the addresses for that account.
Both are valid.

src/views/AddAddressLedger.vue
danimoh: Yes, several things in LedgerApi should be renamed now.
I noted that on my list, but didn't want to do it in this PR.

src/views/AddAddressLedger.vue
danimoh: Should be ok, as these variables don't need to be renamed anymore

src/views/AddAddressLedger.vue
danimoh: I added it to `SignupLedger` and `AddAddressLedger` for consistency to other success screens.
But now I see, that actually only `LoginSuccess` is emitting that event.
I guess it can be removed there as well?

src/views/SignupLedger.vue
sisou: I mean, because what was `accounts` before is now `addresses`, so to simple _translate_, it would be `Fetching Adresses`. But I have no opinion on what's _better_.

PR #184, opened 2019-03-31
Check for invalid account in rpc api already and allow logout of deleted accounts

danimoh:
Waiting for https://github.com/nimiq/accounts/pull/183 to be merged first.

- Check for existence of account required for a request in `RpcApi` already without the need to load the request view.
- Allow logout from account in Keyguard even if it was deleted (e.g by the browser) in the accounts manager database

### DIFF
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -138,7 +138,7 @@ export default class RpcApi {
                     request = RequestParser.parse(arg, state, requestType) || undefined;
                     this._staticStore.request = request;
                 } catch (error) {
-                    state.reply(ResponseStatus.ERROR, error);
+                    this.reject(error);
                     return;
                 }
 
@@ -154,7 +154,7 @@ export default class RpcApi {
                     account = await WalletStore.Instance.get((request as ParsedSimpleRequest).walletId);
                     if (!account && requestType !== RequestType.LOGOUT) {
                         // no account found although it's required
-                        state.reply(ResponseStatus.ERROR, new Error('Account ID not found'));
+                        this.reject(new Error('Account ID not found'));
                         return;
                     }
                 }
@@ -226,7 +226,7 @@ export default class RpcApi {
                 this._recoverState(state);
 
                 if (error.message === ERROR_CANCELED) {
-                    this._staticStore.rpcState!.reply(ResponseStatus.ERROR, error);
+                    this.reject(error);
                     return;
                 }
 
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -147,17 +147,22 @@ export default class RpcApi {
                     hasRequest: !!this._staticStore.request,
                 });
 
-                if (location.pathname !== '/') {
-                    // Don't jump back to request's initial view on reload when navigated to a subsequent view.
-                    // E.g. if the user switches from Checkout to Import, don't jump back to Checkout on reload.
-                    return;
-                }
-
                 let account;
                 // Simply testing if the property exists (with `'walletId' in request`) is not enough,
                 // as `undefined` also counts as existing.
                 if (request && (request as ParsedSimpleRequest).walletId) {
                     account = await WalletStore.Instance.get((request as ParsedSimpleRequest).walletId);
+                    if (!account) {
+                        // no account found although it's required
+                        state.reply(ResponseStatus.ERROR, new Error('Account ID not found'));
+                        return;
+                    }
+                }
+
+                if (location.pathname !== '/') {
+                    // Don't jump back to request's initial view on reload when navigated to a subsequent view.
+                    // E.g. if the user switches from Checkout to Import, don't jump back to Checkout on reload.
+                    return;
                 }
 
                 if (account && account.type === WalletType.LEDGER
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -152,7 +152,7 @@ export default class RpcApi {
                 // as `undefined` also counts as existing.
                 if (request && (request as ParsedSimpleRequest).walletId) {
                     account = await WalletStore.Instance.get((request as ParsedSimpleRequest).walletId);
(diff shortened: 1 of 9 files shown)


src/lib/RpcApi.ts
```diff
@@ -147,17 +147,22 @@ export default class RpcApi {
                     hasRequest: !!this._staticStore.request,
                 });
 
+                let account;
+                if (request && 'walletId' in request) {
+                    account = await WalletStore.Instance.get((request as ParsedSimpleRequest).walletId);
+                    if (!account && requestType !== RequestType.LOGOUT) {
+                        // no account found although it's required
+                        state.reply(ResponseStatus.ERROR, new Error('Account ID not found'));
+                        return;
```
sisou: If handling this error here, the handling can be removed in every other request component, right?

sisou: Additionally, the error text should be made into a constant in `Constants.ts`.

src/views/Logout.vue
```diff
@@ -13,15 +13,11 @@ export default class Logout extends Vue {
...
-            this.$rpc.reject(new Error('Account ID not found'));
-            return;
-        }
 
         const request: KeyguardClient.RemoveKeyRequest = {
             appName: this.request.appName,
             keyId: this.request.walletId,
-            keyLabel: wallet.label,
+            keyLabel: !!wallet ? wallet.label : 'I want to log out', // the label to type in Keyguard to confirm logout
         };
```
sisou: Double `!!` required?

src/lib/RpcApi.ts
danimoh: To remove it, a `!` non-null assertion needs to be added to the `walletStore.get` calls to make typescript know that it exists:
```ts
const wallet = (await WalletStore.Instance.get(this.request.walletId))!;
```
Also in weird circumstances it might still make sense, e.g. if a developer opens the popup pointing to `accounts.nimiq.com/sign-transaction` but with an onboarding request, the `SignTransaction` view can refuse to do anything due to non existing wallet.

That's why I left it in. But now that you bring it up again, I also think it should be removed.
I'll leave the error message as a string then, as it then only has one occurrence.

src/views/Logout.vue
danimoh: Not required.
Maybe nicer?
Shows that it's used as a boolean condition here.
But could equally be removed.

src/lib/RpcApi.ts
sisou: Hmm, that scenario with an invalid request is a valid argument. Maybe the existance of required properties in the request should be checked in the `RequestParser.ts`?

danimoh: I removed the checks in the request views and created an issue for the request validation: https://github.com/nimiq/accounts/issues/190

src/views/ExportSuccess.vue
```diff
@@ -28,14 +28,8 @@ export default class ExportSuccess extends Vue {
...
 
-    @Getter private findWallet!: (id: string) => WalletInfo | undefined;
-
     private async mounted() {
-        const wallet = this.findWallet(this.request.walletId);
-        if (!wallet) {
-            this.$rpc.reject(new Error('walletId not found'));
-            return;
-        }
+        const wallet = (await WalletStore.Instance.get(this.request.walletId))!;
```
danimoh: Removed the use of `findWallet` here for consistency to other request views.

src/lib/RpcApi.ts
```diff
@@ -147,17 +147,22 @@ export default class RpcApi {
                     hasRequest: !!this._staticStore.request,
                 });
 
+                let account;
+                if (request && 'walletId' in request) {
+                    account = await WalletStore.Instance.get((request as ParsedSimpleRequest).walletId);
+                    if (!account && requestType !== RequestType.LOGOUT) {
+                        // no account found although it's required
+                        state.reply(ResponseStatus.ERROR, new Error('Account ID not found'));
```
sisou: This should probably use `this.reject(new Error(...))` instead.

danimoh: Shouldn't it then be changed here
https://github.com/nimiq/accounts/blob/ddae45bd6b1c27165c9ceb22841c8ebdc455b158/src/lib/RpcApi.ts#L141
and here
https://github.com/nimiq/accounts/blob/ddae45bd6b1c27165c9ceb22841c8ebdc455b158/src/lib/RpcApi.ts#L227
too?

sisou: Yes I guess it can. Those were forgotten when we added the reject method.

src/views/Logout.vue
Bettelstab: What about I18n of that label? Wouldn't it be better to let keyguard handle that?

src/lib/RpcApi.ts
```diff
@@ -147,17 +147,22 @@ export default class RpcApi {
                     hasRequest: !!this._staticStore.request,
                 });
 
+                let account;
+                if (request && 'walletId' in request) {
```
sisou: As I noticed in my PR (https://github.com/nimiq/accounts/pull/197/files#diff-5354875441714f08e8db0466207385fdL157), the check for `walletId in request` also returns true when `request.walletId` is `undefined` (which happens when the RequestParser does `request.walletId = publicRequest.accountId`. If `accountId` does not exist, `walletId` gets set on the request object, but as `undefined`. I thus propose to change this check in the same way as in the above linked diff.

src/views/Logout.vue
danimoh: i18n could also be handled here in AM once we add i18n to the AM

src/lib/RpcApi.ts
danimoh: Alright :+1:

src/views/Logout.vue
Bettelstab: Yeah ok, still it seems strange to me to misuse the request parameter like that.

PR #185, opened 2019-04-01
Skip getting balance for imported legacy account

### LINKED ISSUE #175: Skip WalletInfoCollector for legacy accounts
Currently, the WalletInfoCollector runs when importing only a legacy account. This is not necessary.

sisou:
This prevents having to wait for balance when the only possible account is already known. The balance is currently not used anywhere in Accounts, is also checked separately from the Safe and is checked again in Accounts before Checkout.

Resolves #175.

### DIFF
--- a/src/lib/WalletInfoCollector.ts
+++ b/src/lib/WalletInfoCollector.ts
@@ -52,7 +52,7 @@ export default class WalletInfoCollector {
         const walletInfo = await WalletInfoCollector._getWalletInfoInstance(walletId, walletType);
 
         // Add initial accounts to the walletInfo
-        let initialAccountsPromise;
+        let initialAccountsPromise = Promise.resolve();
         if (initialAccounts && initialAccounts.length > 0) {
             WalletInfoCollector._addAccounts(walletInfo, initialAccounts, undefined);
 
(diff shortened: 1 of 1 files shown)


src/lib/WalletInfoCollector.ts
```diff
@@ -55,11 +55,14 @@ export default class WalletInfoCollector {
         let initialAccountsPromise;
```
danimoh:
```suggestion
         let initialAccountsPromise = Promise.resolve();
```

src/lib/WalletInfoCollector.ts
```diff
@@ -55,11 +55,14 @@ export default class WalletInfoCollector {
...
-            });
+
+            if (walletType !== WalletType.LEGACY) {
+                // fetch balances and update again
+                initialAccountsPromise = WalletInfoCollector._getBalances(initialAccounts).then(async (balances) => {
+                    WalletInfoCollector._addAccounts(walletInfo, initialAccounts, balances);
+                    onUpdate(walletInfo, await derivedAccountsPromise);
+                });
+            }
         } else {
```
danimoh:
```suggestion
```

src/lib/WalletInfoCollector.ts
```diff
@@ -55,11 +55,14 @@ export default class WalletInfoCollector {
...
             initialAccountsPromise = Promise.resolve();
```
danimoh:
```suggestion
```

sisou:
Ha, I actually had already done that on my WIP branch, but did not add it to this branch. Nice!

PR #187, opened 2019-04-01
Small window reload fixes

danimoh:
Small cleanup and fix for reload behavior:
- Don't transform `Uint8Array`s in stored `staticStore.keyguardRequest` as the `RpcApi` is able to handle them directly.
- Don't jump back to request's initial view on reload when navigated to a subsequent view. Relevant when user navigates from checkout to onboarding and reloads, from checkout to SignTransactionLedger and reloads or from onboarding to SignupLedger and reloads.

### DIFF
--- a/src/components/Network.vue
+++ b/src/components/Network.vue
@@ -36,9 +36,6 @@ class Network extends Vue {
         return client.connectPico(addresses);
     }
 
-    // Uint8Arrays cannot be stored in SessionStorage, thus the stored request has arrays instead and is
-    // thus not exactly the type KSignTransactionRequest. Thus all potential Uint8Arrays are converted
-    // into Nimiq.SerialBuffers (sender, recipient, data).
     public async prepareTx(
         keyguardRequest: KeyguardClient.SignTransactionRequest,
         keyguardResult: KeyguardClient.SignTransactionResult,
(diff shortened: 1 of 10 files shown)


src/lib/StaticStore.ts
```diff
@@ -16,7 +17,7 @@ export class StaticStore {
     public request?: ParsedRpcRequest;
     public kind?: RequestType;
     public rpcState?: RpcState;
-    public keyguardRequest?: any;
+    public keyguardRequest?: KeyguardRequest;
```
sisou: Can optionally also `import KeyguardClient from '@nimiq/keyguard-client` and then use `KeyguardClient.Request` as the type. I think this is how we do it in other places.

src/components/Network.vue
```diff
@@ -53,15 +50,15 @@ class Network extends Vue {
             || keyguardRequest.recipientType !== Nimiq.Account.Type.BASIC
         ) {
             tx = new Nimiq.ExtendedTransaction(
-                new Nimiq.Address(new Nimiq.SerialBuffer(keyguardRequest.sender)),
+                new Nimiq.Address(keyguardRequest.sender),
```
danimoh: Oh cool, I didn't know the Nimiq primitives accept generic `Uint8Arrays`.
I somehow always thought they expect `SerialBuffer`s. :+1:

src/lib/StaticStore.ts
danimoh: Yes, but importing whole namespaces is actually not considered best practice.
I'd leave it like this.

src/components/Network.vue
sisou: I thought that too back then, that's why it used serialbuffers nearly everywhere.

Turns out, the constructor accepts (and actually needs) an uint8array and the `unserialize` methods require a serialbuffer.

danimoh: `SerialBuffer` extends `Uint8Array`, thus it was also ok before, just not necessary.

sisou: Yes, it _works_. But internally, the primitives use Uint8Arrays, and when constructing one with a SerialBuffer, it assigns it to `this._obj`, which then throws a type-mismatch in tests :D

So using Uint8Arrays in the constructor is the _correct_ way.

Issue #170, opened 2019-03-18
New rename UX


Bettelstab:
- [x] Account Preview
- [x] Button color should be lightBlue

sisou:
Resolved by #215.

PR #189, opened 2019-04-04
KEY_NOT_FOUND error page and hiding keyMissing accounts

### LINKED ISSUE #121: Handle key missing flag on account/walletInfo

sisou:
Accounts marked with `keyMissing = true` will be able to be also managed in the revamped management screen (#170). Until then, they will simply be hidden from apps.

Resolves #121.

### DIFF
--- a/src/lib/WalletInfoCollector.ts
+++ b/src/lib/WalletInfoCollector.ts
@@ -201,7 +201,10 @@ export default class WalletInfoCollector {
         }
 
         const existingWalletInfo = await WalletStore.Instance.get(walletId);
-        if (existingWalletInfo) return existingWalletInfo;
+        if (existingWalletInfo) {
+            existingWalletInfo.keyMissing = false;
+            return existingWalletInfo;
+        }
 
         const label = walletType === WalletType.LEGACY
             ? ACCOUNT_DEFAULT_LABEL_LEGACY
--- a/src/views/ErrorHandler.vue
+++ b/src/views/ErrorHandler.vue
@@ -3,11 +3,12 @@
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator';
 import { State } from 'vuex-class';
-import { Static } from '../lib/StaticStore';
+import staticStore, { Static } from '../lib/StaticStore';
 import { ParsedRpcRequest, ParsedSimpleRequest, RequestType } from '../lib/RequestTypes';
 import { Errors } from '@nimiq/keyguard-client';
 import { WalletStore } from '../lib/WalletStore';
 import KeyguardClient from '@nimiq/keyguard-client';
+import { DEFAULT_KEY_PATH } from '../lib/Constants';
 
 @Component
 export default class ErrorHandler extends Vue {
@@ -19,7 +20,7 @@ export default class ErrorHandler extends Vue {
         if (!(this.keyguardResult instanceof Error)) return;
         if (this.requestSpecificErrors()) return;
         if (this.keyguardResult.message === Errors.Messages.KEY_NOT_FOUND) {
-            let walletId;
+            let walletId: string;
             if ((this.request as ParsedSimpleRequest).walletId) {
                 // The walletId is already in the Accounts request
                 walletId = (this.request as ParsedSimpleRequest).walletId;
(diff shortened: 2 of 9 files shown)


src/lib/WalletInfoCollector.ts
```diff
@@ -149,7 +149,10 @@ export default class WalletInfoCollector {
 
     private static async _getWalletInfoInstance(walletId: string, walletType: WalletType): Promise<WalletInfo> {
         const existingWalletInfo = await WalletStore.Instance.get(walletId);
-        if (existingWalletInfo) return existingWalletInfo;
+        if (existingWalletInfo) {
+            existingWalletInfo.keyMissing = false;
```
Bettelstab: Why do we know this here?
Does this happen after login?

sisou: Yes, exactly. Just in case you are importing a previously deleted key.

However, this brings me to another problem that I think I found: because keys in the KG are incremental, the ID that gets returned from import is another one than the one requested and not found. So I don't think that the collector even finds the previous meta data here...

This needs comparing of requested and new IDs and an update of the store (and thus also an update of the store in the apps, like Safe). All because of the incremental key IDs...

sisou: Or rather, the collector _does_ find the correct wallet info, but fails to update the ID currently, so on the next request the keyguard ID is again not found.

sisou: And this change if ID bubbles up to the apps using the AM, as now they also need to know about and handle the if change.

danimoh: Not nice indeed...
However, instead of changing existing records with the new key id, the accounts manager could also map the not anymore existing id to the new id in all requests.
On reimports, the labels of the old account info can be copied to the new one.
This way, at least apps like the safe wouldn't need to handle this and can continue submitting requests with the old id.

Both approaches result in sort of a mess though.

src/views/ErrorHandler.vue
```diff
@@ -1,25 +1,52 @@
...
+                :state="state"
+                :message="message"
+                :mainAction="mainActionText"
+                :alternativeAction="alternativeActionText"
+                @main-action="mainActionHandler"
+                @alternative-action="alternativeActionHandler"
+                />
+        </SmallPage>
+    </div>
+</template>
```
danimoh: Probably need to remove the empty template from classes inheriting from `ErrorHandler` (`CheckoutErrorHandler`, `LoginErrorHandler`, `SignMessageErrorHandler`, `SignupErrorHandler`) as it probably overwrites this one.

src/views/ErrorHandler.vue
```diff
@@ -1,25 +1,52 @@
...
     @Static protected keyguardRequest?: KeyguardClient.Request;
     @State protected keyguardResult!: Error;
 
+    private title = '';
+    private state = Loader.State.ERROR;
+    private message = '';
+    private mainActionText = '';
+    private alternativeActionText = '';
+    private mainActionHandlerDict: {[error: string]: (event: Event) => void} = {};
+    private alternativeActionHandlerDict: {[error: string]: (event: Event) => void} = {};
```
danimoh: Instead of using the dictionaries here, couldn't you simply directly assign the handler to properties `mainActionHandler` and `alternativeActionHandler`?

src/lib/WalletInfoCollector.ts
Bettelstab: It would be possible to add the old keyId to the login request. In keyguard we would check if the id is still free, and if yes, save the reimported key with the given id.

src/views/ErrorHandler.vue
sisou: Yes, I think so. I had it differently first because I feared Vue's reactivity will break it. But then had to change it to regular dicts and it works. So now I can simplify it again :D

sisou:
@Bettelstab that is taken care of by this simple line: https://github.com/nimiq/accounts/pull/189/files#diff-cfb8e44b6757fd740446e9e4486b73c1R129

Bettelstab:
> @Bettelstab that is taken care of by this simple line: https://github.com/nimiq/accounts/pull/189/files#diff-cfb8e44b6757fd740446e9e4486b73c1R129

Alright. :)

src/lib/WalletInfoCollector.ts
sisou: This is resolved with #205.

PR #195, opened 2019-04-05
Request parsing: required and default properties, extraData as string

### LINKED ISSUE #190: Add request validation in request parser
E.g. check whether all required properties like `accountId` are present

sisou:
Resolves #190.

- Enforce minimum set of required request properties
- Set defaults for optional properties to clean up setting fallbacks throughout the code
- Allow `extraData` for checkout and signTransaction to be a string

This actually fixes a bug in vue-components/PaymentInfoLine, where it displayed 'NaN NIM' when `fee` was omitted in the checkout request.

### DIFF
--- a/src/lib/RequestParser.ts
+++ b/src/lib/RequestParser.ts
@@ -25,6 +25,10 @@ export class RequestParser {
         switch (requestType) {
             case RequestType.SIGN_TRANSACTION:
                 const signTransactionRequest = request as SignTransactionRequest;
+
+                if (!signTransactionRequest.value) throw new Error ('value is required');
+                if (!signTransactionRequest.validityStartHeight) throw new Error('validityStartHeight is required');
+
                 return {
                     kind: RequestType.SIGN_TRANSACTION,
                     appName: signTransactionRequest.appName,
@@ -40,6 +44,8 @@ export class RequestParser {
                 } as ParsedSignTransactionRequest;
             case RequestType.CHECKOUT:
                 const checkoutRequest = request as CheckoutRequest;
+
+                if (!checkoutRequest.value) throw new Error ('value is required');
                 if (checkoutRequest.shopLogoUrl && new URL(checkoutRequest.shopLogoUrl).origin !== state.origin) {
                     throw new Error(
                         'shopLogoUrl must have same origin as caller website. Image at ' +
@@ -79,6 +85,9 @@ export class RequestParser {
             case RequestType.LOGOUT:
             case RequestType.ADD_ADDRESS:
                 const simpleRequest = request as SimpleRequest;
+
+                if (!simpleRequest.accountId) throw new Error('accountId is required');
+
                 return {
                     kind: requestType,
                     appName: simpleRequest.appName,
@@ -86,6 +95,9 @@ export class RequestParser {
                 } as ParsedSimpleRequest;
             case RequestType.RENAME:
                 const renameRequest = request as RenameRequest;
+
+                if (!renameRequest.accountId) throw new Error('accountId is required');
+
                 return {
                     kind: RequestType.RENAME,
                     appName: renameRequest.appName,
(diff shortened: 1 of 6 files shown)


src/lib/RequestParser.ts
```diff
@@ -19,27 +19,36 @@ import {
...
 } from './RequestTypes';
+import { Utf8Tools } from '@nimiq/utils';
 
 export class RequestParser {
     public static parse(request: RpcRequest, state: State, requestType: RequestType): ParsedRpcRequest | null {
         switch (requestType) {
             case RequestType.SIGN_TRANSACTION:
                 const signTransactionRequest = request as SignTransactionRequest;
+
+                if (!signTransactionRequest.value) throw new Error ('value is required');
```
nibhar:
```suggestion
                if (!signTransactionRequest.value) throw new Error('value is required');
```

src/lib/RequestParser.ts
```diff
@@ -19,27 +19,36 @@ import {
...
+                    data: typeof signTransactionRequest.extraData === 'string'
+                        ? Utf8Tools.stringToUtf8ByteArray(signTransactionRequest.extraData)
+                        : signTransactionRequest.extraData || new Uint8Array(0),
+                    flags: signTransactionRequest.flags || Nimiq.Transaction.Flag.NONE,
                     validityStartHeight: signTransactionRequest.validityStartHeight,
                 } as ParsedSignTransactionRequest;
             case RequestType.CHECKOUT:
                 const checkoutRequest = request as CheckoutRequest;
+
+                if (!checkoutRequest.value) throw new Error ('value is required');
```
nibhar:
```suggestion
                if (!checkoutRequest.value) throw new Error('value is required');
```

src/lib/RequestParser.ts
```diff
@@ -19,27 +19,36 @@ import {
...
 
 export class RequestParser {
     public static parse(request: RpcRequest, state: State, requestType: RequestType): ParsedRpcRequest | null {
         switch (requestType) {
             case RequestType.SIGN_TRANSACTION:
                 const signTransactionRequest = request as SignTransactionRequest;
+
+                if (!signTransactionRequest.value) throw new Error ('value is required');
+                if (!signTransactionRequest.validityStartHeight) throw new Error('validityStartHeight is required');
+
```
nibhar: there are more non-optional parameters to this request. Why check for `value` and `validityStartHeight` but not for the others? I think the `sender` and `recipient` will throw if not valid addresses, but `appName` and `accountId` should be checked.

src/lib/RequestParser.ts
```diff
@@ -79,13 +90,19 @@ export class RequestParser {
             case RequestType.LOGOUT:
             case RequestType.ADD_ADDRESS:
                 const simpleRequest = request as SimpleRequest;
+
+                if (!simpleRequest.accountId) throw new Error('accountId is required');
```
nibhar: I think this should also be enforced for `SIGN_TRANSACTION` and `CHECKOUT`. Probably `appName`, too.

sisou: `CHECKOUT` does not use an `accountId`, for `SIGN_TRANSACTION` and `SIGN_MESSAGE` is was removed earlier in #197.

Only question is if we should check for an `appName`.

src/lib/RequestParser.ts
sisou: The `accountId` requirement was removed in #197.

nibhar: I see. Still leaves `appName`, which can not be undefined.

src/lib/RequestParser.ts
sisou: I added a check for the appName to the top, so it is checked for all requests.

src/lib/RequestParser.ts
sisou: https://github.com/nimiq/accounts/pull/195#discussion_r273036120

PR #197, opened 2019-04-05
Remove accountId param from signTransaction and signMessage requests

### LINKED ISSUE #194: Find accountId of sender/signer address automatically
For signTransaction and signMessage, where currently the `accountId` is required in addition to the `sender`/`signer` address, the Accounts Manager should find the corresponding `accountId` automatically if the address is supplied with the request.

This works around the problem, that we don't want to give out the `accountId` with an address in the `chooseAddress` request.

This automatic finding should **only** be enabled for `signTransaction` and `signMessage`, as all other requests requiring an `accountId` are considered _management functions_ and should only be allowed when the `manageAccount` method/permission is available and a user can grant an app access to a whole account.

sisou:
Resolves #194.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,10 +278,9 @@ interface Address {
 
 The `signTransaction()` method is similar to checkout, but provides a different
 UI to the user. The main difference to `checkout()` is that it requires the
-request to already include the sender's account ID and address as `accountId` and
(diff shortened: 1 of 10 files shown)


sisou:
Ha, good question, I didn't even think about removing accountId. Hmm, there should be no problem removing it...

PR #198, opened 2019-04-06
Use Account[] for list and migration results

sisou:
Instead of `WalletInfoEntry`, output the public `Account` type (as array) for `list()` and `migrate()` methods (latter of which results is currently not used in Safe, as migration is done via redirect and the new list of accounts is loaded with list() on return page load).

This is required, as the `WalletInfoEntry` does not contain contracts' user-friendly address anywhere, only addresses in serialized form (the addresses are a map with the user-friendly address as key, but the contracts are not). Because we have no serialized-to-user-friendly-address methods in the Safe (no Nimiq lib, and I also don't want to add a helper for this), we need another result type for contracts.

It was thus fitting to change the whole result to the public result type also returned by `signup()` and `login()` etc.

Adaptions for the Safe are in https://github.com/nimiq/safe/pull/81.

### DIFF
--- a/src/lib/ContractInfo.ts
+++ b/src/lib/ContractInfo.ts
@@ -13,6 +13,7 @@ export class ContractInfoHelper {
         }
     }
 
+    // Used in iframe
     public static objectToContractType(o: ContractInfoEntry): VestingContract | HashedTimeLockedContract {
         switch (o.type) {
             case 1 /* Nimiq.Account.Type.VESTING */:
@@ -39,9 +40,10 @@ export class VestingContractInfo {
         );
     }
 
+    // Used in iframe
     public static objectToContractType(o: VestingContractInfoEntry): VestingContract {
         return {
-            type: o.type,
+            type: 1 /* Nimiq.Account.Type.VESTING */,
             label: o.label,
             address: AddressUtils.toUserFriendlyAddress(o.address),
             owner: AddressUtils.toUserFriendlyAddress(o.owner),
@@ -152,9 +154,10 @@ export class HashedTimeLockedContractInfo {
         );
     }
 
+    // Used in iframe
     public static objectToContractType(o: HashedTimeLockedContractInfoEntry): HashedTimeLockedContract {
         return {
-            type: Nimiq.Account.Type.HTLC,
+            type: 2 /* Nimiq.Account.Type.HTLC */,
             label: o.label,
             address: AddressUtils.toUserFriendlyAddress(o.address),
             sender: AddressUtils.toUserFriendlyAddress(o.sender),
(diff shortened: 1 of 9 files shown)


src/lib/ContractInfo.ts
```diff
@@ -11,6 +12,17 @@ export class ContractInfoHelper {
             default: throw new Error('Unknown contract type: ' + o.type);
         }
     }
+
+    public static objectToContractType(o: ContractInfoEntry): VestingContract | HashedTimeLockedContract {
+        switch (o.type) {
+            case 1 /* Nimiq.Account.Type.VESTING */:
```
nibhar: Why does this not use `Nimiq.Account.Type.VESTING` and `Nimiq.Account.Type.HTLC`?

sisou: Because Nimiq lib is not loaded in iframe.

sisou: Although I think the compiler actually compiles the variable out and leaves the number in, because it's shorter.

src/lib/ContractInfo.ts
```diff
@@ -27,6 +39,19 @@ export class VestingContractInfo {
         );
     }
 
+    public static objectToContractType(o: VestingContractInfoEntry): VestingContract {
+        return {
+            type: o.type,
```
nibhar: For `HashedTimeLockedContractInfo` this value is set explicitly in the corresponding function.

src/lib/ContractInfo.ts
sisou: Correct for 157. The others should not be executed, as the class is only used as static in the iframe. But maybe that just makes my point in saying that the compiler removes the variables and only uses the numbers anyway, so should we maybe leave the variables in?

nibhar: If we can, we should use the constants. If we can not, then we should be consistent about it I think and use the values with the corresponding comments everywhere.

sisou: We cannot use constants. They stay in the built output and throw a `Nimiq is not defined` error.

nibhar: I would then say remove the other occurrences of `Nimiq.*`, too.

sisou: No, I say we should only use the values with commented constants where necessary, not automatically everywhere.

nibhar: Personally I find it somewhat strange to have both versions in the same file.

sisou: I opened https://github.com/nimiq/accounts/issues/200 to track this problem.

PR #205, opened 2019-04-10
Deterministic wallet ids

sisou:
Adapt to https://github.com/nimiq/keyguard-next/pull/255 and use the full `keyId` for internal communication with the Keyguard, while using a shortened 6 byte hex `id` externally (could also be shortened to 4 bytes, but needs to be changed for Ledger as well then -> broader discussion).

The `keyId` is stored with a `WalletInfo` and in the DB with the `WalletInfoEntry`, but not in the cookie, as it's not an external property and thus not required when listing accounts in the iframe.

Ledgers also use no `keyId` and set it to an empty string, as it's never used (we can also declare another constant string here, but it's not necessary).

Both the IndexedDB and the cookie `VERSION` are increased to delete previously stored accounts, just like in the Keyguard.


src/lib/WalletStore.ts
```diff
@@ -1,12 +1,30 @@
...
+
+        const existingIds = wallets.map((wallet) => wallet.id);
+        const keyIdBytes = Nimiq.BufferUtils.fromBase64(keyId);
+
+        for (let i = 0; i <= (32 - WalletStore.WALLET_ID_LENGTH); i++) {
+            const id = Nimiq.BufferUtils.toHex(keyIdBytes.subarray(i, i + WalletStore.WALLET_ID_LENGTH));
+            if (existingIds.indexOf(id) === -1) return id;
+        }
+
+        throw new Error('Could not find an available wallet ID');
```
Bettelstab: While this line might never be hit, we can do better - e.g. increment until it works.

src/lib/WalletInfoCollector.ts
```diff
@@ -30,6 +30,7 @@ export type WalletCollectionResult = {
 export default class WalletInfoCollector {
     public static async collectWalletInfo(
         walletType: WalletType,
+        keyId: string,
         walletId?: string,
```
Bettelstab: If I'm not mistaken, this parameter is not needed anymore as it is always undefined.

nibhar: I think so, too.

src/lib/WalletStore.ts
nibhar: or hash the id again in a `do { } while( )`

PR #206, opened 2019-04-10
Hide contracts in signMessage

sisou:
Add a filter to remove contracts from `processedWallets`.

Also fix wording from 'account' to 'address'.

### DIFF
--- a/src/views/SignMessage.vue
+++ b/src/views/SignMessage.vue
@@ -9,10 +9,10 @@
                 <span class="origin">{{ originDomain }}</span>
             </div>
 
-            <h1 class="nq-h1">Choose an account to sign</h1>
+            <h1 class="nq-h1">Choose an address to sign</h1>
 
             <AccountSelector
-                :wallets="processedWallets"
+                :wallets="processedWallets | withoutContracts"
                 @account-selected="setAccount"
                 @login="goToOnboarding"/>
         </SmallPage>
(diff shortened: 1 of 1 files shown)


src/views/SignMessage.vue
```diff
@@ -39,7 +39,17 @@ import { Utf8Tools } from '@nimiq/utils';
 import { ERROR_CANCELED } from '@/lib/Constants';
 import { State as RpcState } from '@nimiq/rpc';
 
-@Component({components: {SmallPage, AccountSelector}})
+@Component({
+    components: {SmallPage, AccountSelector},
+    filters: {
+        withoutContracts(wallets: WalletInfo[]): WalletInfo[] {
+            return wallets.map((wallet) => {
+                wallet.contracts = [];
```
danimoh: I think this has side effects that might even change content in the store.
Please use a computed property that returns a filtered copy or introduce a flag `withoutContracts` on the `AccountSelector`.

Bettelstab: Yep, I fear Daniel's right. I was too quick approving this, sorry. My first thought was actually a flag in AccountsSelector. But at whatever place, Object.assign to the rescue :)

sisou: Yes, I was aware that the filter affects the source object, but since contracts are never/nowhere used in during the SignMessage account selection and the Vuex store does have no effect on the IndexedDB, I thought this to be a non-problem.

PR #207, opened 2019-04-10
Network: pico adaptions

danimoh:
Requires a new network client package to be released first


Bettelstab:
...and a redeployment of network, too.

PR #208, opened 2019-04-11
Export splitting and success Screen

nibhar:
This PR updates the export request to include ways to skipFile and skipWords. It also updates the success screen of export depending on what actually was exported. 

Needs  nimiq/keyguard-next#263


demos/Demo.ts
```diff
@@ -325,11 +328,34 @@ class Demo {
...
-        } as SimpleRequest;
+    public async exportFile(accountId: string) {
+        try {
+            const result = await this._accountsClient.export({
+                appName: 'Accounts Demos',
+                accountId,
+                fileOnly: true,
+            });
+            console.log('Keyguard result', result);
+            document.querySelector('#result').textContent = 'Export sucessful';
```
sisou: The message should be conditional to the actual result flag. Because it can be `false`.

demos/Demo.ts
```diff
@@ -325,11 +328,34 @@ class Demo {
...
+
+    public async exportWords(accountId: string) {
+        try {
+            const result = await this._accountsClient.export({
+                appName: 'Accounts Demos',
+                accountId,
+                wordsOnly: true,
+            });
+            console.log('Keyguard result', result);
+            document.querySelector('#result').textContent = 'Export sucessful';
```
sisou: Same here.

src/lib/RequestParser.ts
```diff
@@ -99,6 +100,16 @@ export class RequestParser {
                     appName: simpleRequest.appName,
                     walletId: simpleRequest.accountId,
                 } as ParsedSimpleRequest;
+
```
sisou: There are no empty lines between any of the other cases.

src/views/Export.vue
```diff
@@ -2,22 +2,24 @@
 
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator';
-import { ParsedSimpleRequest } from '../lib/RequestTypes';
-import { SimpleRequest } from '@nimiq/keyguard-client';
+import { ParsedExportRequest } from '../lib/RequestTypes';
+import { ExportRequest } from '@nimiq/keyguard-client';
```
sisou: To reduce confusion over request type names, combine this:

```suggestion
import { KeyguardClient } from '@nimiq/keyguard-client';
```

src/views/Export.vue
```diff
@@ -2,22 +2,24 @@
...
 @Component
 export default class Export extends Vue {
-    @Static private request!: ParsedSimpleRequest;
+    @Static private request!: ParsedExportRequest;
 
     public async created() {
         const wallet = (await WalletStore.Instance.get(this.request.walletId))!;
 
-        const request: SimpleRequest = {
+        const request: ExportRequest = {
```
sisou: With this:

```suggestion
        const request: KeyguardClient.ExportRequest = {
```

src/views/Export.vue
Bettelstab: I guess without braces?

sisou: Ah, yes. Importing the namespace, not the property. I was confused because we call the namespace `KeyguardClient` in the other files, but the package also exports a named `KeyguardClient` :smile:.

PR #211, opened 2019-04-12
Fix vesting contract check for imported ledger

danimoh:


src/lib/WalletInfoCollector.ts
```diff
@@ -59,9 +59,25 @@ export default class WalletInfoCollector {
...
+            const potentialVestingOwner = walletType === WalletType.LEGACY
+                ? initialAccounts[0] // for legacy wallets we get the only account as initialAccounts from the Keyguard
+                : (await derivedAccountsPromise)[0]; // for ledger we check the first derived account
+            const contracts = await WalletInfoCollector._getGenesisVestingContractsForAddress(
+                Nimiq.Address.fromUserFriendlyAddress(potentialVestingOwner.address));
+            WalletInfoCollector._addContracts(walletInfo, contracts);
+            if (contracts.length > 0
+                && !initialAccounts.some((account) => account.address === potentialVestingOwner.address)) {
+                // make sure a ledger vesting owner account get's added even if it has no balance or transaction history
+                initialAccounts.push(potentialVestingOwner);
```
Bettelstab: Just for my understanding: For ledger, there is not such a thing as a default address (= first address) which is always added regardless of balance / history?

danimoh: Exactly. For Ledger, the user can pick his first address

PR #214, opened 2019-04-16
Allow passing a sender to checkout

### LINKED ISSUE #212: Allow optional sender parameter to checkout
Allow submitting an optional `sender` address to the checkout request to skip address selection if the address exists.

The implementation would follow the sign-message request, which already implements this behavior.

The reason for this feature is that apps which already have access to a user's address to work with, can currently just use sign-transaction, which would not display the app url and logo for tx to the app. Some devs would like this flexibility.

sisou:
When the `sender` address exists and has enough balance, the user is directly forwarded to the Keyguard for signing.

When the address does either not exist or does not have sufficient balance, the user is either displayed the address selector list (default), or - in case `forceSender = true` - the request is rejected with an exception.

Resolves #212.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -110,9 +110,9 @@ const store: StoreOptions<RootState> = {
         findWallet: (state) => (id: string): WalletInfo | undefined => {
             return state.wallets.find((wallet) => wallet.id === id);
         },
-        findWalletByAddress: (state) => (address: string): WalletInfo | undefined => {
+        findWalletByAddress: (state) => (address: string, includeContracts: boolean): WalletInfo | undefined => {
             const foundWallet = state.wallets.find((wallet) => wallet.accounts.has(address));
-            if (foundWallet) return foundWallet;
+            if (foundWallet || !includeContracts) return foundWallet;
             return state.wallets.find((wallet) => wallet.contracts.some((contract) => {
                 return contract.address.toUserFriendlyAddress() === address;
             }));
--- a/src/views/Checkout.vue
+++ b/src/views/Checkout.vue
@@ -99,39 +99,41 @@ export default class Checkout extends Vue {
 
         await this.handleOnboardingResult();
 
-        if (this.wallets.length === 0) this.goToOnboarding(true);
-        else {
-            const balances = await this.getBalances();
+        if (this.wallets.length === 0) {
+            this.goToOnboarding(true);
+            return;
+        }
 
-            // Handle optional sender address included in the request
-            if (this.request.sender) {
-                let errorMsg = '';
-                // Check if the address exists
-                const senderAddress = this.request.sender.toUserFriendlyAddress();
-                const wallet = this.findWalletByAddress(senderAddress, true);
-                if (wallet) {
-                    // Check if that address has enough balance
-                    const balance = balances.get(senderAddress);
-                    if (balance && balance >= this.minBalance) {
-                        // Forward to Keyguard, skipping account selection
-                        this.setAccountOrContract(wallet.id, senderAddress);
-                        return;
-                    } else {
-                        errorMsg = 'Address does not have sufficient balance';
-                    }
-                } else {
-                    errorMsg = 'Address not found';
-                }
+        const balances = await this.getBalances();
 
-                if (this.request.forceSender) {
-                    this.$rpc.reject(new Error(errorMsg));
+        // Handle optional sender address included in the request
+        if (this.request.sender) {
+            let errorMsg = '';
+            // Check if the address exists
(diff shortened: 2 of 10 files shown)


src/views/SignMessage.vue
```diff
@@ -13,7 +13,7 @@
 
             <AccountSelector
                 :wallets="processedWallets | withoutContracts"
-                @account-selected="setAccount"
+                @account-selected="accountSelected"
```
sisou: This is just a refactoring of the method name.

src/store.ts
```diff
@@ -110,9 +110,9 @@ const store: StoreOptions<RootState> = {
         findWallet: (state) => (id: string): WalletInfo | undefined => {
             return state.wallets.find((wallet) => wallet.id === id);
         },
-        findWalletByAddress: (state) => (address: string): WalletInfo | undefined => {
+        findWalletByAddress: (state) => (address: string, includeContracts: boolean): WalletInfo | undefined => {
```
sisou: I made the new argument non-optional on purpose, so you _have to_ think about if you want contracts or not. Should reduce dev errors.

src/views/Checkout.vue
```diff
@@ -98,7 +100,38 @@ export default class Checkout extends Vue {
         await this.handleOnboardingResult();
 
         if (this.wallets.length === 0) this.goToOnboarding(true);
```
Bettelstab: Please add `{ }`

src/views/SignMessage.vue
```diff
@@ -70,17 +70,17 @@ export default class SignMessage extends Vue {
         await this.handleOnboardingResult();
 
         if (this.request.signer) {
-            const wallet = this.findWalletByAddress(this.request.signer.toUserFriendlyAddress());
+            const wallet = this.findWalletByAddress(this.request.signer.toUserFriendlyAddress(), false);
             if (wallet) {
-                this.setAccount(wallet.id, this.request.signer.toUserFriendlyAddress(), true);
+                this.accountSelected(wallet.id, this.request.signer.toUserFriendlyAddress(), true);
```
Bettelstab: not a fan, but ok

src/views/Checkout.vue
```diff
@@ -98,7 +100,38 @@ export default class Checkout extends Vue {
...
+            await this.getBalances();
+
+            // Handle optional sender address included in the request
+            if (this.request.sender) {
+                let errorMsg = '';
+                // Check if the address exists
+                const wallet = this.findWalletByAddress(this.request.sender.toUserFriendlyAddress(), true);
+                if (wallet) {
+                    // Check if that address has enough balance
+                    const accountOrContract = wallet.accounts.get(this.request.sender.toUserFriendlyAddress())
```
Bettelstab: would be cool to have a getter for this, so we could skip the wallet lookup.

sisou: I don't understand. The wallet lookup in three lines above is essentially the check if the passed address exists at all and is even required, because we need the wallet id/keyId.

The only thing that could be simplified here is not looking up the AccountInfo/ContractInfo object again, but to look up the balance in the balance map used in `getBalances()`. I can do this by returning that map from the function.

Bettelstab: Ah, we need the wallet id, right. Fine then.

sisou:
TODO, so I don't forget: add `forceSigner` to `signMessage` request also.

src/views/Checkout.vue
```diff
@@ -97,74 +99,109 @@ export default class Checkout extends Vue {
...
+            let errorMsg = '';
+            // Check if the address exists
+            const senderAddress = this.request.sender.toUserFriendlyAddress();
+            const wallet = this.findWalletByAddress(senderAddress, true);
+            if (wallet) {
+                // Check if that address has enough balance
+                const balance = balances.get(senderAddress);
+                if (balance && Nimiq.Policy.coinsToSatoshis(balance) >= this.minBalance) {
+                    // Forward to Keyguard, skipping account selection
+                    this.setAccountOrContract(wallet.id, senderAddress);
```
danimoh: Does this work on reload?
Wouldn't the `getBalances` be instant in that case and `setAccountOrContract` not execute due to `this.height` being 0?

sisou: There is not really a possibility for a reload, because if this line gets executed, the user is directly forwarded to the Keyguard. No GUI is displayed in the AM.

sisou: Furthermore, `getBalances()` calls `this.onHeadChange(cache)` which sets `this.height` anyway.

PR #216, opened 2019-04-23
Ledger Sign Transaction / Checkout flow

danimoh:


sisou:
All these ellipses mean, that your comment message titles are too long. Commit messages should be 60 characters max in the first line, then have one empty line and then have a body of text.

image

danimoh:
Well, I like commit message to be expressive instead of just messages like "Address review comments". I usually add which class is affected which serves as short synopsis.

sisou:
> Well, I like commit message to be expressive instead of just messages like "Address review comments". I usually add which class is affected which serves as short synopsis.

Yes, you can do that, but not in the first line of the commit message. For a longer explanation you should use the commit message body, which is separated from a short title by an empty line.

src/lib/LedgerApi.ts
```diff
@@ -326,39 +323,57 @@ class LedgerApi {
...
+
+        if (transaction.extraData && transaction.extraData.length !== 0
+            || senderType !== Nimiq.Account.Type.BASIC
+            || recipientType !== Nimiq.Account.Type.BASIC
+            || flags !== Nimiq.Transaction.Flag.NONE
+        ) {
+            const extraData = transaction.extraData ? transaction.extraData : new Uint8Array(0);
+            nimiqTx = new Nimiq.ExtendedTransaction(senderAddress, senderType, transaction.recipient, recipientType,
+                transaction.value - fee, fee, transaction.validityStartHeight, flags, extraData, undefined,
+                networkId);
```
sisou: `networkId` can be left out, it defaults to the currently initialized GenesisConfig anyway.

src/lib/LedgerApi.ts
```diff
@@ -326,39 +323,57 @@ class LedgerApi {
...
         } else {
-            nimiqTx = new Nimiq.BasicTransaction(senderPubKey, recipient, value, fee, transaction.validityStartHeight,
-                undefined, networkId);
+            nimiqTx = new Nimiq.BasicTransaction(senderPubKey, transaction.recipient, transaction.value, fee,
+                transaction.validityStartHeight, undefined, networkId);
```
sisou: `networkId` can be left out, it defaults to the currently initialized GenesisConfig anyway.

src/lib/RpcApi.ts
```diff
@@ -150,11 +155,24 @@ export default class RpcApi {
...
-                        this.reject(new Error('Account ID not found'));
+                if (request) {
+                    let accountRequired;
+                    if ((request as ParsedSimpleRequest).walletId) {
+                        accountRequired = true;
+                        account = await WalletStore.Instance.get((request as ParsedSimpleRequest).walletId);
+                    } else if (requestType === RequestType.SIGN_TRANSACTION) {
+                        accountRequired = true;
+                        const address = (request as ParsedSignTransactionRequest).sender;
+                        account = this._store.getters.findWalletByAddress(address.toUserFriendlyAddress());
```
sisou: Typescript sadly does not throw an error here, because `this._store.getters` is `any`, but the `findWalletByAddress()` function expects _two_ arguments, the second being whether to include contracts in the search or not. Right now, you do not, but you must.

src/lib/RpcApi.ts
```diff
@@ -150,11 +155,24 @@ export default class RpcApi {
...
+                    } else if (requestType === RequestType.SIGN_MESSAGE) {
+                        accountRequired = false; // Sign message allows user to select an account
+                        const address = (request as ParsedSignMessageRequest).signer;
+                        if (address) {
+                            account = this._store.getters.findWalletByAddress(address.toUserFriendlyAddress());
+                        }
+                    }
```
sisou: The `CHECKOUT` flow is missing here, similar to sign_message, where a `sender` is not required but can be included (added with #214, after you started this branch).

Additionally, checkout has a `forceSender` option, which should throw an error when the passed `sender` does not exist. Thus I would say to set `accountRequired = request.forceSender` to trigger the correct error behavior.

src/views/Checkout.vue
```diff
@@ -217,44 +215,57 @@ export default class Checkout extends Vue {
...
-            flags: this.request.flags,
-        };
-
-        staticStore.keyguardRequest = request;
+        this.proceedToTransactionSigning(walletInfo, accountInfo!, contractInfo);
+    }
 
-        const client = this.$rpc.createKeyguardClient();
-        client.signTransaction(request);
+    private proceedToTransactionSigning(walletInfo: WalletInfo, accountInfo: AccountInfo, contractInfo?: ContractInfo) {
```
sisou: I removed this redundant function splitting in #214 as well: https://github.com/nimiq/accounts/pull/214/files#diff-10870cb1266d47895e2ffe5e572207e0L189

src/views/SignTransactionLedger.vue
```diff
@@ -0,0 +1,408 @@
...
+
+    private created() {
+        let senderAccountInfo;
+        let senderAddressInfo;
+        if (this.request.kind === RequestType.SIGN_TRANSACTION) {
+            // direct sign transaction request invocation
+            const signTransactionRequest = this.request as ParsedSignTransactionRequest;
+            const senderAddress = signTransactionRequest.sender.toUserFriendlyAddress();
+            senderAccountInfo = this.findWalletByAddress(senderAddress)!;
+            senderAddressInfo = senderAccountInfo.accounts.get(senderAddress)!;
```
sisou: If the sender is a contract, need to also search contracts.

src/views/SignTransactionLedger.vue
```diff
@@ -0,0 +1,408 @@
...
+export default class SignTransactionLedger extends Vue {
+    private static readonly State = {
+        OVERVIEW: 'overview',
+        SENDING_TRANSACTION: 'sending-transaction',
+        FINISHED: 'finished',
+    };
+
+    @Static private rpcState!: RpcState;
+    @Static private request!: ParsedSignTransactionRequest | ParsedCheckoutRequest;
+    @Getter private findWalletByAddress!: (address: string) => WalletInfo | undefined;
```
sisou: This function as a second mandatory parameter `includeContracts: boolean` as of #214.

src/views/SignTransactionLedger.vue
```diff
@@ -0,0 +1,408 @@
...
+            validityStartHeight = blockchainHeight + 1
+                - TX_VALIDITY_WINDOW
+                + checkoutRequest.validityDuration;
+        } else {
+            // this case get's rejected in created
+            return;
+        }
+
+        const senderAccountInfo = this.findWalletByAddress(this.senderDetails.address)!;
+        const senderAddressInfo = senderAccountInfo.accounts.get(this.senderDetails.address)!;
```
sisou: As above, you need to consider contracts here as well.

src/views/SignTransactionLedger.vue
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    private get transactionFee() {
+        return Nimiq.Policy.satoshisToCoins(this.request.fee || 0);
+    }
+
+    private get transactionData() {
+        return this.request.data && this.request.data.length > 0
+            ? Utf8Tools.utf8ByteArrayToString(this.request.data)
+            : null;
```
sisou: You might want to do a check here, if the data is even valid utf8, and otherwise display it as HEX.

src/lib/LedgerApi.ts
```diff
@@ -326,39 +323,57 @@ class LedgerApi {
...
-                Nimiq.Transaction.Flag.NONE, transaction.extraData, undefined, networkId);
+        const flags = transaction.flags !== undefined
+            ? transaction.flags
+            : Nimiq.Transaction.Flag.NONE;
+        const fee = transaction.fee || 0;
+
+        if (transaction.extraData && transaction.extraData.length !== 0
+            || senderType !== Nimiq.Account.Type.BASIC
+            || recipientType !== Nimiq.Account.Type.BASIC
+            || flags !== Nimiq.Transaction.Flag.NONE
```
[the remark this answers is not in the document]
Bettelstab: I would not do that. You would have to add a comment explaining what happens there, while this code is self-explanatory.

src/components/LedgerUi.vue
```diff
@@ -154,7 +156,7 @@ class LedgerUi extends Vue {
                 this._showInstructions('', 'The connected Ledger is not the one this account belongs to.');
                 break;
             case LedgerApi.ErrorType.REQUEST_ASSERTION_FAILED:
-                this._showInstructions('Request failed', error.message);
+                this._showInstructions('Request failed', `${this.small ? 'Request failed: ' : ''}${error.message}`);
```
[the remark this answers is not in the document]
danimoh: Yes, as on small layout the title is not shown. Therefore I add the `'Request Failed: '` failed here as I don't know whether the error message that I'm getting itself is self explanatory enough.

src/components/Network.vue
```diff
@@ -236,6 +268,8 @@ class Network extends Vue {
         if (networkClient.consensusState === 'syncing') this.$emit(Network.Events.CONSENSUS_SYNCING);
         else if (networkClient.consensusState === 'established') this.$emit(Network.Events.CONSENSUS_ESTABLISHED);
         else if (networkClient.consensusState === 'lost') this.$emit(Network.Events.CONSENSUS_LOST);
+        Network._hadInitialConsensus = Network._hadInitialConsensus || networkClient.consensusState === 'established';
+        this.consensusEstablished = Network._hadInitialConsensus && networkClient.consensusState !== 'lost';
```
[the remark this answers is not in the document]
danimoh: Not really. It's only to set the initial state consistently to already existing Network component instances. For a component that was created along the network client, `consensusEstablished` is true after consensus, including during consensus syncing,  as long as no `consensus-lost` event happens.
Only during consensus syncing state after consensus lost but before consensus established, the value of `consensusEstablished` is not accurate. I can circumvent that by resetting `_hadInitialConsensus` (which should then be renamed) to false on consensus lost.

src/lib/Constants.ts
```diff
@@ -28,9 +28,10 @@ export const LEGACY_GROUPING_ACCOUNT_ID = 'LEGACY';
 export const LEGACY_GROUPING_ACCOUNT_LABEL = 'Single Accounts';
 
 // Networks
-export const NETWORK_TEST = 'test';
-export const NETWORK_MAIN = 'main';
-export const NETWORK_DEV = 'dev';
+type NetworkType = 'test' | 'main' | 'dev';
+export const NETWORK_TEST: NetworkType = 'test';
+export const NETWORK_MAIN: NetworkType = 'main';
+export const NETWORK_DEV: NetworkType = 'dev';
```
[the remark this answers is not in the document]
danimoh: Without `NetworkType` typescript assumes `string` as type which is not compatible with the expected `'main' | 'test' | 'bounty' | 'dev'` in `LedgerApi`.

src/lib/LedgerApi.ts
```diff
@@ -66,20 +66,17 @@ const LedgerJs: LedgerJs = require('@nimiq/ledgerjs/ledgerjs-nimiq.min.js');
...
-    value: number;
-    fee: number;
+    sender: Nimiq.Address;
+    senderType?: Nimiq.Account.Type;
+    recipient: Nimiq.Address;
+    recipientType?: Nimiq.Account.Type;
+    value: number; // In Luna
+    fee?: number;
     validityStartHeight: number;
     network: 'main' | 'test' | 'bounty' | 'dev';
```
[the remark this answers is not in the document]
danimoh: Hum, yeah, could do that. However, I plan to move `LedgerApi` into its own repository with the ledger base library someday. Therefore it's better to not add that additional dependency here.
If the types are not consistent at some point, typescript is gonna be able to catch that.

src/lib/LedgerApi.ts
```diff
@@ -326,39 +323,57 @@ class LedgerApi {
...
-            !== senderAddress.replace(/ /g, '')) {
+        const senderAddress = senderPubKey.toAddress();
+        if (!senderAddress.equals(transaction.sender)) {
             LedgerApi._throwError(LedgerApi.ErrorType.REQUEST_ASSERTION_FAILED,
-                'Sender Address doesn\'t match this ledger account', request);
+                'Sender Address doesn\'t match this ledger address', request);
         }
+        const senderType = transaction.senderType !== undefined
+            ? transaction.senderType
+            : Nimiq.Account.Type.BASIC;
```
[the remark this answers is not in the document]
danimoh: These are written on purpose in a more expressive way. Writing `transaction.senderType || Nimiq.Account.Type.BASIC` requires the additional knowledge that not another type than `Nimiq.Account.Type.BASIC` is 0.

src/lib/LedgerApi.ts
```diff
@@ -326,39 +323,57 @@ class LedgerApi {
...
-        const recipient = Nimiq.Address.fromUserFriendlyAddress(transaction.recipient);
-        const value = Nimiq.Policy.coinsToSatoshis(transaction.value);
-        const fee = Nimiq.Policy.coinsToSatoshis(transaction.fee);
-        if (transaction.extraData && transaction.extraData.length !== 0) {
-            nimiqTx = new Nimiq.ExtendedTransaction(senderPubKey.toAddress(), Nimiq.Account.Type.BASIC,
-                recipient, Nimiq.Account.Type.BASIC, value - fee, fee, transaction.validityStartHeight,
-                Nimiq.Transaction.Flag.NONE, transaction.extraData, undefined, networkId);
+        const flags = transaction.flags !== undefined
+            ? transaction.flags
+            : Nimiq.Transaction.Flag.NONE;
```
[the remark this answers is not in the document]
danimoh: Also written on purpose like that.

src/lib/LedgerApi.ts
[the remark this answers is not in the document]
danimoh: Yes, this was also on purpose.
Only the `!== 0` might be debatable.

src/lib/LedgerApi.ts
danimoh: The `LegderApi` is meant to be usable by other devs outside of the accounts manager some day.
Therefore the support for specifying the network.

src/lib/LedgerApi.ts
danimoh: The `LegderApi` is meant to be usable by other devs outside of the accounts manager some day.
Therefore the support for specifying the network.

src/views/Checkout.vue
danimoh: I reintroduced again as I liked it now that the method now has become longer.

danimoh: As the `setAccountOrContract` method got shorter after some refactoring, thus I merged the methods again.

danimoh: As the `setAccountOrContract` method got shorter after some refactoring, thus I merged the methods again.

src/views/SignTransactionLedger.vue
danimoh: See https://github.com/nimiq/accounts/issues/232

danimoh:
Ready for re-review.
Note that while the frontend now does support sending transactions from vesting contracts, the actual nimiq ledger app does not yet.

src/lib/LedgerApi.ts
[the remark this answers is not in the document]
sisou: So if I pass in `null` as `transaction.senderType`, then this is not going to catch it, and furthermore 351 this would lead to the transaction becoming an extended one, although not necessary. Same for `recipientType` and `flags`.

src/lib/LedgerApi.ts
sisou: Haha, but above you are setting it to fixed `networkId = genesisConfig.NETWORK_ID`, the same as leaving it out here would do. It's not even configurable.

src/lib/LedgerApi.ts
```diff
@@ -328,39 +325,53 @@ class LedgerApi {
...
+        const senderType = transaction.senderType !== undefined
+            ? transaction.senderType
+            : Nimiq.Account.Type.BASIC;
+
+        const recipientType = transaction.recipientType !== undefined
+            ? transaction.recipientType
+            : Nimiq.Account.Type.BASIC;
+
         const genesisConfig = Nimiq.GenesisConfig.CONFIGS[transaction.network];
         const networkId = genesisConfig.NETWORK_ID;
```
sisou: If not using typescript, this will throw (`genesisConfig is undefined`) when `transaction.network` is not one of the possible strings or is left out. Is that on purpose?

src/lib/LedgerApi.ts
sisou: Then I'd say you need to make sure, as commented above, that `networkId` actually falls back to the configured GenesisConfig, instead of throwing an exception.

src/views/Checkout.vue
```diff
@@ -202,59 +202,59 @@ export default class Checkout extends Vue {
...
+                    validityStartHeight,
+                    data: this.request.data,
+                    flags: this.request.flags,
+                };
+
+                staticStore.keyguardRequest = request;
+                const client = this.$rpc.createKeyguardClient();
+                client.signTransaction(request);
+                return;
+            case WalletType.LEDGER:
```
sisou: You could move the LEDGER case to the top, as it's very short and would be better for visibility, I think.

src/lib/LedgerApi.ts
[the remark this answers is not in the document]
danimoh: Typescript is will complain about the incompatible input value / type `null`.

PR #217, opened 2019-04-23
Ledger Logout

danimoh:


src/views/LogoutLedger.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+    <div class="container">
+        <SmallPage>
+            <PageHeader>Mind the loss of settings</PageHeader>
+            <PageBody>
+                <div class="warning nq-text nq-red">
+                    Logging out will delete custom settings and labels. Contacts are not affected.
```
sisou: Clarify that it deletes stuff _for this account only_, and I would leave out the contacts part, because contacts are not yet stored in Accounts.
```suggestion
                    Logging out will delete custom settings and labels for this account.
```

src/views/LogoutLedger.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+    <div class="container">
+        <SmallPage>
+            <PageHeader>Mind the loss of settings</PageHeader>
```
sisou: Is setting a good word? Please check with Julian, because for me "addresses and labels" are not _settings_ really.

src/views/LogoutLedger.vue
```diff
@@ -0,0 +1,96 @@
...
+import { ERROR_CANCELED } from '../lib/Constants';
+import Loader from '../components/Loader.vue';
+
+@Component({components: {SmallPage, PageHeader, PageBody, Loader}})
+export default class LogoutLedger extends Vue {
+    @Static private request!: ParsedSimpleRequest;
+
+    private confirmedLogout: boolean = false;
+
+    private async _logOut() {
```
sisou: Methods in typescript really do not need to have underscores, and AFAIK no other component/view is using underscored private methods.

src/views/LogoutLedger.vue
danimoh: All the wording is coming from Julian / Tammo.

src/views/LogoutLedger.vue
danimoh: The contacts part is needed as it's not clear to the user wheere contacts are stored or related to.
Tammo and Julian were actually surprised that the contacts persist.
It doesn't make a lot of sense for third party apps without contacts though.

src/views/LogoutLedger.vue
danimoh: We had a decision to mark private methods with underscores.

src/views/LogoutLedger.vue
danimoh: I changed it to only display the note about contacts when coming from the safe

PR #222, opened 2019-05-03
Ledger chrome 72+ support

danimoh:


danimoh:
One version to rule them all ;)

Also fixes Brave support which also got broken at some point.

PR #228, opened 2019-05-06
Handle absence of request

Bettelstab:
Introduces a new config parameter `redirectTarget`, where the user is redirected if the rpc server doesn't receive a request. Needs https://github.com/nimiq/rpc/pull/17 to be merged and released first.

I also moved the config files from root to `src/config` to emphasise that they take effect at build-time and not at run-time and to be consistent with keyguard.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,6 +4,7 @@ import router from './router';
 import store from './store';
 import staticStore from '@/lib/StaticStore';
 import RpcApi from '@/lib/RpcApi';
+import Config from 'config';
 import VueRaven from 'vue-raven'; // Sentry.io SDK
 
 // Register service worker if necessary (and possible).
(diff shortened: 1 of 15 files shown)


src/main.ts
```diff
@@ -15,6 +16,11 @@ if ('serviceWorker' in navigator) {
     });
 }
 
+// Redirect to Safe (default) if referrer is empty
+if (!document.referrer) {
```
sisou: Is this necessary here, when the RpcServer already handles it?

This might break the Accounts Manager for browser extensions again, as they do not submit a referrer, I think. Also, does this not trigger for popups?

Bettelstab: It does not trigger for popups. But you're right, client timeout will happen anyway. Removed it.

PR #229, opened 2019-05-07
Enable automatic Account retrieval in case login returns multiple Accounts

### LINKED ISSUE #147: Handle KeyResult with multiple entries
If import by words is ambiguous, we have to decide which one to keep.

nibhar:
This PR updates `LoginSuccess` and `WalletInfoCollector`. 

Single returned Accounts still get added as usual. 
For multiple returned Accounts, if there are some which have either a balance or a transaction history, only those will be added. Otherwise none of the Accounts was previously used and then only the newest AccountType (BIP39) will be added.

The accounts which did not got added, will also correctly be released in Keyguard.

needs nimiq/keyguard-next#297

resolves #147

### DIFF
--- a/src/lib/WalletInfoCollector.ts
+++ b/src/lib/WalletInfoCollector.ts
@@ -1,5 +1,5 @@
 import { NetworkClient } from '@nimiq/network-client';
-import { KeyguardClient } from '@nimiq/keyguard-client';
+import { KeyguardClient, SimpleResult as KeyguardSimpleResult } from '@nimiq/keyguard-client';
 import { AccountInfo } from '@/lib/AccountInfo';
 import { WalletStore } from '@/lib/WalletStore';
 import { WalletInfo, WalletType } from '@/lib/WalletInfo';
@@ -25,9 +25,8 @@ type BasicAccountInfo = {
 export type WalletCollectionResult = {
     walletInfo: WalletInfo,
     receiptsError?: Error, // if there is an incomplete result due to failed requestTxReceipts requests
-    releaseKey: (removeKey: boolean) => void,
+    releaseKey: (removeKey?: boolean) => Promise<KeyguardSimpleResult | void>,
     hasActivity: boolean, // whether the wallet has a transaction history or a balance or owns a contract
-
 };
 
 export default class WalletInfoCollector {
@@ -105,9 +104,7 @@ export default class WalletInfoCollector {
 
             return {
                 walletInfo,
-                releaseKey: async (removeKey) => {
-                    await WalletInfoCollector._keyguardClient!.releaseKey(keyId, removeKey);
-                },
+                releaseKey: (removeKey?) => WalletInfoCollector._keyguardClient!.releaseKey(keyId, removeKey),
                 hasActivity,
             };
         }
@@ -187,13 +184,15 @@ export default class WalletInfoCollector {
             LedgerApi.currentRequest.cancel();
         }
 
+        const releaseKey = walletType === WalletType.LEDGER
+            ? (removeKey?: boolean) => removeKey ? Promise.reject('Can\'t remove Ledger key') : Promise.resolve()
+            : (removeKey?: boolean) => WalletInfoCollector._keyguardClient!.releaseKey(keyId, removeKey);
+
         await initialAccountsPromise; // make sure initial accounts balances are updated
         return {
             walletInfo,
             receiptsError,
-            releaseKey: (removeKey) => {
-                WalletInfoCollector._keyguardClient!.releaseKey(keyId, removeKey);
-            },
+            releaseKey,
             hasActivity,
         };
     }
--- a/src/lib/WalletInfoCollector.ts
+++ b/src/lib/WalletInfoCollector.ts
@@ -26,7 +26,8 @@ export type WalletCollectionResult = {
     walletInfo: WalletInfo,
     receiptsError?: Error, // if there is an incomplete result due to failed requestTxReceipts requests
     releaseKey: (removeKey: boolean) => void,
-    hasHistoryOrBalance: boolean, // if the wallet has a transaction history or a balance
+    hasActivity: boolean, // whether the wallet has a transaction history or a balance or owns a contract
+
 };
 
 export default class WalletInfoCollector {
@@ -39,10 +40,10 @@ export default class WalletInfoCollector {
         checkLegacyActivity = true,
     ): Promise<WalletCollectionResult> {
         // Kick off loading dependencies
-        WalletInfoCollector._initializeDependencies(walletType);
+        WalletInfoCollector._initializeDependencies(walletType, checkLegacyActivity);
 
         // track activity of the wallet
-        let hasHistoryOrBalance = false;
+        let hasActivity = false;
         // Kick off first round of account derivation
         let startIndex = 0;
         let derivedAccountsPromise: Promise<BasicAccountInfo[]>;
@@ -65,11 +66,13 @@ export default class WalletInfoCollector {
             const contracts = await WalletInfoCollector._getGenesisVestingContractsForAddress(
                 Nimiq.Address.fromUserFriendlyAddress(potentialVestingOwner.address));
             WalletInfoCollector._addContracts(walletInfo, contracts);
-            if (contracts.length > 0
-                && !initialAccounts.some((account) => account.address === potentialVestingOwner.address)) {
-                // make sure a ledger vesting owner account get's added even if it has no balance or transaction history
-                initialAccounts.push(potentialVestingOwner);
-                hasHistoryOrBalance = true;
+            if (contracts.length > 0) {
+                if (!initialAccounts.some((account) => account.address === potentialVestingOwner.address)) {
+                    // make sure a ledger vesting owner account get's added
+                    // even if it has no balance or transaction history
+                    initialAccounts.push(potentialVestingOwner);
+                }
+                hasActivity = true;
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/lib/WalletInfoCollector.ts
```diff
@@ -72,20 +77,33 @@ export default class WalletInfoCollector {
...
-            if (walletType !== WalletType.LEGACY) {
-                // fetch balances and update again
-                initialAccountsPromise = WalletInfoCollector._getBalances(initialAccounts).then(async (balances) => {
-                    WalletInfoCollector._addAccounts(walletInfo, initialAccounts, balances);
-                    onUpdate(walletInfo, await derivedAccountsPromise);
-                });
-            }
+            // fetch balances and update again
+            // This step could be skipped for WalletType.LEGACY in case there is only one Account to collect info on.
+            // In that case the WalletInfoCollector needs to wait for consensus unnecessarily.
```
sisou: This is indeed unfavorable. When importing a definite legacy account, one should not have to wait for a consensus.

src/lib/WalletInfoCollector.ts
```diff
@@ -72,20 +77,33 @@ export default class WalletInfoCollector {
...
         }
         onUpdate(walletInfo, await derivedAccountsPromise);
 
-        if (walletType === WalletType.LEGACY) {
+        if (walletType === WalletType.LEGACY && WalletInfoCollector._keyguardClient) {
             // legacy wallets have no derived accounts
             await initialAccountsPromise;
-            return { walletInfo };
+            const address = initialAccounts[0].address;
+            hasHistoryOrBalance = walletInfo.accounts.get(address)!.balance !== undefined &&
```
sisou: Furthermore, this check should only be run when multiple accounts are checked. Otherwise this check is totally unnecessary.

I thus propose to add a boolean flag to the collector initialization, that tells it if it needs to check this at all.

src/views/LoginSuccess.vue
```diff
@@ -36,60 +37,95 @@ export default class LoginSuccess extends Vue {
...
+                        this.retrievalFailed = true;
+                        if (tryCount >= 5) throw e;
+                    }
                 }
+            }),
+        );
+
+        // In case there is only one returned Account it is always added.
+        // WalletType.LEDGER always returns one Account, hence does not need special treatment in the other cases.
+        let condition: (collectionResult: WalletCollectionResult) => boolean = (collectionResult) => true;
```
```suggestion
        let keepWalletDetector: (collectionResult: WalletCollectionResult) => boolean = (collectionResult) => true;
```

src/views/LoginSuccess.vue
```diff
@@ -36,60 +37,97 @@ export default class LoginSuccess extends Vue {
...
+                    } catch (e) {
+                        this.retrievalFailed = true;
+                        if (tryCount >= 5) throw e;
+                    }
                 }
+            }),
+        );
+
+        // In case there is only one returned Account it is always added.
+        // WalletType.LEDGER always returns one Account, hence does not need special treatment in the other cases.
```
danimoh: Ledger does not use this page.

src/views/LoginSuccess.vue
```diff
@@ -36,60 +37,97 @@ export default class LoginSuccess extends Vue {
...
+            } else {
+                // In case of more than one returned account but none saw activity in the past
+                // look for the BIP39 account and add it while discarding the others.
+                keepWalletCondition = (collectionResult) => collectionResult.walletInfo.type === WalletType.BIP39;
             }
-        });
+        }
+
+        await Promise.all (
+            collectionResults.map( async (collectionResult, index) => {
```
danimoh: What do you need the `index` for?

src/views/LoginSuccess.vue
```diff
@@ -36,60 +37,97 @@ export default class LoginSuccess extends Vue {
...
+                accountId: walletInfo.id,
+                label: walletInfo.label,
+                type: walletInfo.type,
+                fileExported: walletInfo.fileExported,
+                wordsExported: walletInfo.wordsExported,
+                addresses: Array.from(walletInfo.accounts.values())
+                    .map((addressInfo) => addressInfo.toAddressType()),
+                contracts: walletInfo.contracts.map((contract) => contract.toContractType()),
+            });
+        }
```
danimoh: Better to do a `this.result = this.walletInfos.map(...)`.

src/views/LoginSuccess.vue
```diff
@@ -18,13 +18,14 @@ import { Component, Vue } from 'vue-property-decorator';
 import { ParsedBasicRequest } from '../lib/RequestTypes';
 import { Account } from '../lib/PublicRequestTypes';
 import { State } from 'vuex-class';
-import { WalletInfo } from '../lib/WalletInfo';
+import { WalletInfo, WalletInfoEntry, WalletType } from '../lib/WalletInfo';
```
danimoh: `WalletInfoEntry` seems to be unused?

src/lib/WalletInfoCollector.ts
```diff
@@ -82,10 +89,24 @@ export default class WalletInfoCollector {
...
+            if (checkLegacyActivity) {
+                hasHistoryOrBalance = walletInfo.accounts.get(address)!.balance !== undefined &&
+                    walletInfo.accounts.get(address)!.balance !== 0 ||
+                    (await NetworkClient.Instance.requestTransactionReceipts(address)).length > 0;
+            }
+
+            return {
+                walletInfo,
+                releaseKey: (removeKey) => {
+                    WalletInfoCollector._keyguardClient!.releaseKey(keyId, removeKey);
```
danimoh: Should be awaited

client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nimiq/accounts-client",
-  "version": "0.3.1",
+  "version": "0.4.0-beta.1",
```
danimoh: To be changed on merge

src/lib/WalletInfoCollector.ts
```diff
@@ -64,6 +69,7 @@ export default class WalletInfoCollector {
                 && !initialAccounts.some((account) => account.address === potentialVestingOwner.address)) {
                 // make sure a ledger vesting owner account get's added even if it has no balance or transaction history
                 initialAccounts.push(potentialVestingOwner);
+                hasHistoryOrBalance = true;
```
danimoh: This line should not depend on the condition `contracts.length > 0`.
As the second condition is not fulfilled for legacy accounts.

Also here one sees, that `hasHistoryOrBalance` is not a well chosen name.

src/lib/WalletInfoCollector.ts
```diff
@@ -25,6 +25,8 @@ type BasicAccountInfo = {
 export type WalletCollectionResult = {
     walletInfo: WalletInfo,
     receiptsError?: Error, // if there is an incomplete result due to failed requestTxReceipts requests
+    releaseKey: (removeKey: boolean) => void,
+    hasHistoryOrBalance: boolean, // if the wallet has a transaction history or a balance
```
danimoh: Should this rather be called `hasActivity` (or `wasUsed` /  `hasUsage`) to match the `checkLegacyActivity`?
```suggestion
    hasActivity: boolean, // whether the wallet has a transaction history or a balance or owns a contract
```

src/lib/WalletInfoCollector.ts
```diff
@@ -34,10 +36,13 @@ export default class WalletInfoCollector {
         initialAccounts: BasicAccountInfo[] = [],
         // tslint:disable-next-line:no-empty
         onUpdate: (walletInfo: WalletInfo, currentlyCheckedAccounts: BasicAccountInfo[]) => void = () => {},
+        checkLegacyActivity = true,
```
danimoh: Maybe:
```suggestion
        checkActivity = walletType !== WalletType.LEGACY,
```

src/lib/WalletInfoCollector.ts
```diff
@@ -82,10 +89,24 @@ export default class WalletInfoCollector {
         }
         onUpdate(walletInfo, await derivedAccountsPromise);
 
-        if (walletType === WalletType.LEGACY) {
+        if (walletType === WalletType.LEGACY && WalletInfoCollector._keyguardClient) {
```
danimoh: Why are you checking for `_keyguardClient` here?
It's always set by the call to `_initializeDependencies`.
If it's only to avoid typescript complaint use `!` instead.

src/lib/WalletInfoCollector.ts
```diff
@@ -82,10 +89,24 @@ export default class WalletInfoCollector {
...
             // legacy wallets have no derived accounts
             await initialAccountsPromise;
-            return { walletInfo };
+            const address = initialAccounts[0].address;
+
+            if (checkLegacyActivity) {
+                hasHistoryOrBalance = walletInfo.accounts.get(address)!.balance !== undefined &&
```
danimoh: have to use
```ts
hasActivity = hasActivity || ....
```
here to avoid overwriting the information whether the address owns a contract.

src/lib/WalletInfoCollector.ts
```diff
@@ -82,10 +89,24 @@ export default class WalletInfoCollector {
...
+                    walletInfo.accounts.get(address)!.balance !== 0 ||
```
danimoh: Instead of checking for `undefined` and `0` just check for `!!walletInfo.accounts.get(address)!.balance`

src/lib/WalletInfoCollector.ts
```diff
@@ -180,7 +205,7 @@ export default class WalletInfoCollector {
             || NetworkClient.createInstance(Config.networkEndpoint).init();
         WalletInfoCollector._networkInitializationPromise
             .catch(() => delete WalletInfoCollector._networkInitializationPromise);
-        if (walletType === WalletType.BIP39) {
+        if (walletType === WalletType.BIP39 || walletType === WalletType.LEGACY) {
```
danimoh: KeyguardClient only needs to be created if `checkActivity` is true. We can discuss though whether it's actually worth avoiding an unnecessary call.
I'm not entirely sure how expensive this call is, specifically whether it creates a keyguard iframe.

PR #230, opened 2019-05-07
Error reporting blacklist

nibhar:
Errors.InvalidRequest, Account ID not found, Address not found

### DIFF
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -6,7 +6,7 @@ import Router from 'vue-router';
 import { ParsedSimpleRequest, RequestType } from './RequestTypes';
 import { RequestParser } from './RequestParser';
 import { RpcRequest, RpcResult } from './PublicRequestTypes';
-import { KeyguardCommand, KeyguardClient } from '@nimiq/keyguard-client';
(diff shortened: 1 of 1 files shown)


src/lib/RpcApi.ts
```diff
@@ -83,8 +83,9 @@ export default class RpcApi {
     }
 
     public reject(error: Error) {
-        const ignoredErrors = [ ERROR_CANCELED, 'Request aborted' ];
-        if (ignoredErrors.indexOf(error.message) < 0) {
+        const ignoredErrorTypes = [ Errors.Types.INVALID_REQUEST.toString() ];
+        const ignoredErrors = [ ERROR_CANCELED, 'Request aborted', 'Account ID not found', 'Address not found' ];
+        if (!(ignoredErrorTypes.indexOf(error.name) >= 0 || ignoredErrors.indexOf(error.message) >= 0)) {
```
sisou: Now that I think about it, this condition can also be written as
```suggestion
        if (ignoredErrorTypes.indexOf(error.name) < 0 && ignoredErrors.indexOf(error.message) < 0) {
```
It means to only run the capture-exception code, if both ignored-lists are not triggered.

PR #231, opened 2019-05-07
Browser warning for unsupported browsers, private modes etc.

Bettelstab:
Use @nimiq/browser-warning and style the warning.


sisou:
Approved :+1:

PR #234, opened 2019-05-08
RPC: Correctly handle history back from Keyguard

sisou:
Use https://github.com/nimiq/keyguard-next/pull/307 's new `handleHistoryBack` option to correctly react when the user does `history.back()` from the Keyguard.

In nearly all cases the expected behavior is to reject, since there is no user interaction required in the Accounts Manager.


src/lib/RpcApi.ts
```diff
@@ -58,16 +58,24 @@ export default class RpcApi {
...
 
-    public createKeyguardClient() {
+    public createKeyguardClient(handleHistoryBack?: boolean) {
         const localState = this._exportState();
-        const client = new KeyguardClient(Config.keyguardEndpoint, window.location.origin, localState);
+        const client = new KeyguardClient(
+            Config.keyguardEndpoint,
+            window.location.origin,
+            localState,
+            undefined,
```
Bettelstab: please add a comment or an option parameter

package.json
```diff
@@ -15,10 +15,10 @@
     "dependencies": {
         "@nimiq/browser-warning": "^0.3.0",
         "@nimiq/iqons": "^1.4.4",
-        "@nimiq/keyguard-client": "^0.4.3",
+        "@nimiq/keyguard-client": "^0.5.0-beta.2",
         "@nimiq/ledgerjs": "https://github.com/nimiq/ledger-api.git",
         "@nimiq/network-client": "^0.2.0",
-        "@nimiq/rpc": "^0.1.5",
+        "@nimiq/rpc": "^0.2.0-beta.3",
```
nibhar: after the changes are merged reminder to put the correct new version.

PR #235, opened 2019-05-08
Trigger key migration screen when necessary

### LINKED ISSUE #233: Offer backup for all accounts before migration

sisou:
Resolves #233.

When during AM startup it is detected, that no accounts are stored, but the Keyguard has legacy accounts, then the current request is stored as `originalRouteName` and the user is redirected to the migration screen. When returning from an Keyguard-Export request, the user is again redirected to migration.

Only when migration succeeds and accounts are stored in the AM, then everything is back to normal and the orignal request is triggered (if it was not `migrate()` in the first place).

image
image

### DIFF
--- a/src/views/ExportSuccess.vue
+++ b/src/views/ExportSuccess.vue
@@ -14,7 +14,7 @@
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator';
 import { SmallPage, CheckmarkIcon } from '@nimiq/vue-components';
-import { ParsedSimpleRequest } from '../lib/RequestTypes';
+import { ParsedSimpleRequest, RequestType } from '../lib/RequestTypes';
 import { ExportResult } from '../lib/PublicRequestTypes';
 import { State } from 'vuex-class';
 import { Static } from '@/lib/StaticStore';
@@ -26,12 +26,18 @@ import KeyguardClient from '@nimiq/keyguard-client';
 @Component({components: {SmallPage, Loader, CheckmarkIcon}})
 export default class ExportSuccess extends Vue {
     @Static private request!: ParsedSimpleRequest;
+    @Static private originalRouteName?: string;
     @State private keyguardResult!: KeyguardClient.ExportResult;
 
     private loaderState = Loader.State.LOADING;
     private successMessage = '';
 
     private async mounted() {
+        if (this.originalRouteName || this.request.kind === RequestType.MIGRATE) {
+            this.$rpc.routerReplace(RequestType.MIGRATE);
+            return;
+        }
+
         const wallet = (await WalletStore.Instance.get(this.request.walletId))!;
 
         wallet.fileExported = wallet.fileExported || this.keyguardResult.fileExported;
--- a/src/views/ExportSuccess.vue
+++ b/src/views/ExportSuccess.vue
@@ -82,7 +82,7 @@ export default class ExportSuccess extends Vue {
             // We want the RpcApi's reply method to only be invoked when migration is finished.
 
             // Recreate original URL with original query parameters
-            const query = { id: staticStore.rpcState!.id.toString() };
+            const query = { rpcId: staticStore.rpcState!.id.toString() };
             setTimeout(
                 () => this.$router.push({ name: RequestType.MIGRATE, query }), Loader.SUCCESS_REDIRECT_DELAY);
         }
--- a/src/views/ExportSuccess.vue
+++ b/src/views/ExportSuccess.vue
@@ -17,36 +17,39 @@ import { SmallPage, CheckmarkIcon } from '@nimiq/vue-components';
 import { ParsedSimpleRequest, RequestType } from '../lib/RequestTypes';
 import { ExportResult } from '../lib/PublicRequestTypes';
 import { State } from 'vuex-class';
-import { Static } from '@/lib/StaticStore';
+import staticStore, { Static } from '@/lib/StaticStore';
 import Loader from '@/components/Loader.vue';
-import { WalletInfo } from '@/lib/WalletInfo';
 import { WalletStore } from '@/lib/WalletStore';
 import KeyguardClient from '@nimiq/keyguard-client';
 
 @Component({components: {SmallPage, Loader, CheckmarkIcon}})
 export default class ExportSuccess extends Vue {
     @Static private request!: ParsedSimpleRequest;
-    @Static private originalRouteName?: string;
     @State private keyguardResult!: KeyguardClient.ExportResult;
 
     private loaderState = Loader.State.LOADING;
     private successMessage = '';
 
     private async mounted() {
-        if (this.originalRouteName || this.request.kind === RequestType.MIGRATE) {
-            this.$rpc.routerReplace(RequestType.MIGRATE);
-            return;
+        let result: ExportResult;
+
+        // When doing pre-migration exports, or the request is MIGRATE, no wallet will be found.
+        const wallet = this.request.walletId ? await WalletStore.Instance.get(this.request.walletId) : null;
+
+        if (wallet) {
+            wallet.fileExported = wallet.fileExported || this.keyguardResult.fileExported;
+            wallet.wordsExported = wallet.wordsExported || this.keyguardResult.wordsExported;
+
+            result = {
+                fileExported: wallet.fileExported,
+                wordsExported: wallet.wordsExported,
+            };
+        } else {
+            result = {
+                fileExported: this.keyguardResult.fileExported,
+                wordsExported: this.keyguardResult.wordsExported,
+            };
         }
-
-        const wallet = (await WalletStore.Instance.get(this.request.walletId))!;
-
-        wallet.fileExported = wallet.fileExported || this.keyguardResult.fileExported;
-        wallet.wordsExported = wallet.wordsExported || this.keyguardResult.wordsExported;
-
-        const result: ExportResult = {
-            fileExported: wallet.fileExported,
-            wordsExported: wallet.wordsExported,
-        };
 
         if (this.keyguardResult.fileExported) {
             if (this.keyguardResult.wordsExported) {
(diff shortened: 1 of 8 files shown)


src/views/Migrate.vue
```diff
@@ -19,28 +41,45 @@ import { Component, Vue } from 'vue-property-decorator';
...
 
-    public mounted() {
-        this.keyguardClient = this.$rpc.createKeyguardClient();
-        this.run();
+        const client = this.$rpc.createKeyguardClient();
+        client.export(request);
     }
 
-    private async run() {
+    private async runMigration() {
```
nibhar: Why are two methods needed for this? They could be combined into one I think.

sisou: Do you mean `runMigration` and `doMigration`? Because I liked it better to have a small `try {} catch` loop around a single function call, as opposed to around a whole function block.

nibhar: Yes, I thought as much.

src/views/ExportSuccess.vue
```diff
@@ -26,12 +26,18 @@ import KeyguardClient from '@nimiq/keyguard-client';
...
 export default class ExportSuccess extends Vue {
     @Static private request!: ParsedSimpleRequest;
+    @Static private originalRouteName?: string;
     @State private keyguardResult!: KeyguardClient.ExportResult;
 
     private loaderState = Loader.State.LOADING;
     private successMessage = '';
 
     private async mounted() {
+        if (this.originalRouteName || this.request.kind === RequestType.MIGRATE) {
```
danimoh: Add a comment why we're not letting the `RpcApi` handle this as a side result.
And wouldn't one of the conditions be enough?

src/views/Migrate.vue
```diff
@@ -1,6 +1,28 @@
 <template>
     <div class="container pad-bottom">
-        <SmallPage>
+        <SmallPage v-if="!isMigrating">
+            <div class="page-header nq-card-header">
```
danimoh: There is also a `PageHeader` component.
Not a good idea to use that one?

src/views/Migrate.vue
```diff
@@ -19,28 +41,45 @@ import { Component, Vue } from 'vue-property-decorator';
 import { AccountInfo } from '@/lib/AccountInfo';
 import { WalletStore } from '@/lib/WalletStore';
 import { WalletInfo, WalletType } from '@/lib/WalletInfo';
-import { SmallPage } from '@nimiq/vue-components';
+import { SmallPage, PageHeader, PageBody, PageFooter, AccountList } from '@nimiq/vue-components';
```
danimoh: Ok, you're actually importing it even. Either use it or remove it.

src/views/Migrate.vue
```diff
@@ -19,28 +41,45 @@ import { Component, Vue } from 'vue-property-decorator';
...
 import Network from '@/components/Network.vue';
 import Loader from '@/components/Loader.vue';
-import { KeyguardClient } from '@nimiq/keyguard-client';
+import KeyguardClient from '@nimiq/keyguard-client';
```
danimoh:
```suggestion
import { ExportRequest, LegacyKeyInfoObject } from '@nimiq/keyguard-client';
```

src/views/Migrate.vue
```diff
@@ -19,28 +41,45 @@ import { Component, Vue } from 'vue-property-decorator';
...
     private state: Loader.State = Loader.State.LOADING;
     private message: string = '';
 
-    private keyguardClient?: KeyguardClient;
+    private legacyAccounts: AccountInfo[] = [];
+    private isMigrating: boolean = false;
+
+    public async created() {
+        const legacyKeys = await this.$rpc.keyguardClient.listLegacyAccounts();
+        this.legacyAccounts = legacyKeys.map((key) => this.legacyKeyInfoObject2AccountInfo(key));
```
danimoh: As the `legacyKeyInfoObject2AccountInfo` method is only used here, I would inline it.

src/views/Migrate.vue
```diff
@@ -19,28 +41,45 @@ import { Component, Vue } from 'vue-property-decorator';
...
+        const request: KeyguardClient.ExportRequest = {
+            appName: this.request.appName,
+            keyId: address,
+            keyLabel: '',
+        };
 
-    public mounted() {
-        this.keyguardClient = this.$rpc.createKeyguardClient();
-        this.run();
+        const client = this.$rpc.createKeyguardClient();
```
danimoh: Why don't you use `$rpc.keyguardClient`?

src/views/ExportSuccess.vue
```diff
@@ -26,12 +26,18 @@ import KeyguardClient from '@nimiq/keyguard-client';
...
+            this.$rpc.routerReplace(RequestType.MIGRATE);
```
danimoh: I know this is cumbersome, but _ideally_ we would remember which accounts have been successfully exported, to already set the `wordsExported` flags accordingly after migration.

src/views/Migrate.vue
```diff
@@ -19,28 +41,45 @@ import { Component, Vue } from 'vue-property-decorator';
...
-    private keyguardClient?: KeyguardClient;
+    private legacyAccounts: AccountInfo[] = [];
+    private isMigrating: boolean = false;
+
+    public async created() {
+        const legacyKeys = await this.$rpc.keyguardClient.listLegacyAccounts();
+        this.legacyAccounts = legacyKeys.map((key) => this.legacyKeyInfoObject2AccountInfo(key));
+    }
+
+    private accountSelected(walletId: string, address: string) {
```
danimoh: Private method names should start with an underscore.

sisou:
Ideally I could mark this as a draft again, but I can't. So I either close this PR and loose the comments, or you all unsubscribe from notifications for now. You can do that at the bottom of the right sidebar.

src/views/Migrate.vue
sisou: My problem with this is, that `ExportRequest` is not unique, because the Hub also has this type itself. Thus I would have to import it `as KeyguardExportRequest`, which then makes no difference anymore, I think.

src/views/Migrate.vue
sisou: The `$rpc.keyguardClient` can only be used for iframe requests, because it does not get initialized with the `localState` and the `returnUrl`.

sisou:
This PR is now ready again for review. All UI and functionality has been implemented and tested.

src/views/Migrate.vue
sisou: I respectfully disagree. They have the word `private` in front of them and thus TS makes sure they cannot be called from outside.

julianbauer:
One small request: Can we change the text from "... and much more." to "There's much more." Thanks!

sisou:
Back arrow in export is done by https://github.com/nimiq/keyguard-next/pull/343.

src/views/Migrate.vue
danimoh: In general it does make a difference for tree-shake-ability. Importing a whole name space is not good practice and as far as I know will result in all imports being bundled.
Here in the case of the KeyguardClient it doesn't matter.

src/views/Migrate.vue
```diff
@@ -49,56 +214,56 @@ export default class Migrate extends Vue {
...
             await WalletStore.Instance.put(walletInfo);
 
             walletInfos.push(walletInfo);
         }
 
-        this.status = 'Migrating Keyguard...';
-        await this.keyguardClient!.migrateAccountsToKeys();
+        this.status = 'Updating Keyguard...';
+        await this.$rpc.keyguardClient.migrateAccountsToKeys();
+        this.deleteAccountsCache();
```
danimoh: This means this info will not be available on `tryAgain()` anymore.
Is that ok?

src/views/Migrate.vue
```diff
@@ -110,17 +275,312 @@ export default class Migrate extends Vue {
...
+
+    This has a negative visual effect on the check-box, because it has a transition applied
+    which fades it out when it is de-rendered. Because of the DOM-diffing instead of full element
+    replacing, the check-box is rendered fading-out next to the intro page's "Prepare for update"
+    button when clicking on the back arrow.
+
+    These styles here prevent that from happening by force-hiding them.
+
+    If you know a better method to force Vue to throw out an element instead of fading it out,
+    please make yourself known.
```
danimoh: By adding a `key` to the vue elements, this shouldn't happen.

src/views/Migrate.vue
```diff
@@ -1,46 +1,211 @@
...
+                    </div>
+                </div>
+                <div class="topic">
+                    <div class="topic-visual qr-code"><ScanQrCodeIcon/></div>
+                    <p class="topic-text">
+                        Create and scan QR codes to quickly share Addresses.
+                    </p>
+                </div>
+
+                <a href="https://medium.com/nimiq-network" target="_blank" class="nq-link link-read-article">
```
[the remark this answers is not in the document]
sisou: It will when that article is available.

src/views/Migrate.vue
```diff
@@ -49,56 +214,56 @@ export default class Migrate extends Vue {
...
-                account.id,
+                await WalletStore.deriveId(keyInfo.id),
+                keyInfo.id,
                 ACCOUNT_DEFAULT_LABEL_LEGACY,
                 accounts,
                 contracts,
                 WalletType.LEGACY,
+                /* keyMissing */ false,
+                /* fileExported */ false,
+                /* wordsExported */ true,
```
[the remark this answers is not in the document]
sisou: As discussed, when the user clicks "My backups are safe.", then he/she confirms that all backups are safe, and we now agreed to not bother the user afterwards with warnings.

src/views/Migrate.vue
sisou: The `migrateAccountsToKeys()` action is the last action that can go wrong. If that succeeds and the code reaches this line, then nothing else can go wrong, everything is stored and the users gets the success screen.

src/views/Migrate.vue
sisou: I tried `:key`... :man_facepalming:

PR #247, opened 2019-05-11
Cookie space checks

Bettelstab:
Before login / signup / onboarding, checks if there is enough space in cookies to fit additional data. If it turns out that there is not enough space after login and address retrieval, informs the user and throws an error.


src/views/AddAccount.vue
```diff
@@ -1,24 +1,36 @@
-<template></template>
+<template>
+    <NotEnoughCookieSpace v-if='notEnoughCookieSpace'/>
```
nibhar: Couldn't this be its own route?

src/views/LoginSuccess.vue
```diff
@@ -112,10 +117,24 @@ export default class LoginSuccess extends Vue {
...
     }
 
-    private done() {
+    private async done() {
         if (!this.walletInfos.length) throw new Error('WalletInfo not ready.');
 
-        this.result = this.walletInfos.map((walletInfo) => ({
+        if (BrowserDetection.isIOS() || BrowserDetection.isSafari()) {
+            const infoEntries = this.walletInfos.map((walletInfo) => walletInfo.toObject());
+            if (!await CookieHelper.canFitNewWallets(infoEntries)) {
```
nibhar: Seems mostly like a duplicate of the `NotEnoughCookieSpace` component. However there it is classified as an error, here it is a `Loader.State.Warning`. I would say it should rather be the other way around. In case you already added an Account that cannot be stored it should be an error, while attempting to add one should result in a Warning if there is not enough space to do so.

src/views/LoginSuccess.vue
```diff
@@ -112,10 +117,24 @@ export default class LoginSuccess extends Vue {
         this.done();
     }
 
-    private done() {
+    private async done() {
         if (!this.walletInfos.length) throw new Error('WalletInfo not ready.');
 
-        this.result = this.walletInfos.map((walletInfo) => ({
+        if (BrowserDetection.isIOS() || BrowserDetection.isSafari()) {
```
nibhar: Could be done in `mounted()` after the `WalletInfoCollector` did its job, but before the Accounts are added to the indexedDB.

src/lib/CookieJar.ts
```diff
@@ -56,8 +62,41 @@ class CookieJar {
...
-    public static canFitNewWallet(wallet: WalletInfoEntry): boolean {
-        return (this.MAX_COOKIE_SIZE - this.getCookieSize()) >= this.encodeWallet(wallet).length;
+    public static canFitNewWallets(wallets?: WalletInfoEntry[]): boolean {
+        let sizeNeeded = 0;
+
+        if (!wallets) {
+            const dummyAddressHumanReadable = 'NQ86 6D3H 6MVD 2JV4 N77V FNA5 M9BL 2QSP 1P64';
+            const dummyAddress = Nimiq.Address.fromUserFriendlyAddress(dummyAddressHumanReadable);
+            const dummyAddressSerialized = new Uint8Array(dummyAddress.serialize());
+            const dummyWallet = {
```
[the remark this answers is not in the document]
Bettelstab: I like that it adapts to possible changes of encoding. But I can precalculate those values above.

src/views/LoginSuccess.vue
Bettelstab: Made it an error.

Issue #225, opened 2019-05-06
Reject unsupported Ledger requests


danimoh:
Reject unsupported requests like "Export Words" if initiated with Ledger accounts.

PR #250, opened 2019-05-12
Handle unsupported ledger requests

danimoh:
Resolves https://github.com/nimiq/accounts/issues/225

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -50,6 +50,9 @@ const SignMessageSuccess      = () => import(/*webpackChunkName: "sign-message"*
 const SimpleSuccess           = () => import(/*webpackChunkName: "common"*/ './views/SimpleSuccess.vue');
 const ErrorHandler            = () => import(/*webpackChunkName: "common"*/ './views/ErrorHandler.vue');
 
+const ErrorHandlerUnsupportedLedger = () => import(/*webpackChunkName: "unsupported-ledger"*/
+    './views/ErrorHandlerUnsupportedLedger.vue');
+
 Vue.use(Router);
 
 export function keyguardResponseRouter(
@@ -163,6 +166,11 @@ export default new Router({
             name: `${RequestType.EXPORT}-success`,
         },
         {
+            path: `/${RequestType.EXPORT}/ledger`,
+            component: ErrorHandlerUnsupportedLedger,
+            name: `${RequestType.EXPORT}-ledger`,
+        },
+        {
             path: `/${RequestType.CHANGE_PASSWORD}`,
             component: ChangePassword,
             name: RequestType.CHANGE_PASSWORD,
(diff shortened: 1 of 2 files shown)


src/router.ts
```diff
@@ -227,5 +240,10 @@ export default new Router({
             component: SignMessageSuccess,
             name: `${RequestType.SIGN_MESSAGE}-success`,
         },
+        {
+            path: `/${RequestType.SIGN_MESSAGE}/ledger`,
+            component: ErrorHandlerUnsupportedLedger,
+            // not specifying a name here to not trigger automatic routing to this view in RpcApi.ts
```
nibhar: Why is the name omitted here but in the 2 other cases it is not?

src/views/ErrorHandlerUnsupportedLedger.vue
```diff
@@ -0,0 +1,59 @@
...
+        this.state = Loader.State.WARNING;
+        this.title = 'Unsupported Request';
+    }
+
+    private get _message() {
+        switch (this.request.kind) {
+            case RequestType.SIGN_MESSAGE:
+                return 'Message signing is not supported for Ledgers.';
+            case RequestType.EXPORT:
+                return 'Can not export a backup from a Ledger. Please refer to your 24 words backup you created '
```
nibhar:
```suggestion
                return 'Ledger Accounts cannot be exported. Please refer to your 24 words backup you created '
```

src/views/ErrorHandlerUnsupportedLedger.vue
```diff
@@ -0,0 +1,59 @@
...
+        const randomWaitTime = 1000 + Math.round(Math.random() * 2000); // 1-3s
+        await new Promise((resolve) => setTimeout(resolve, randomWaitTime));
+        this.state = Loader.State.WARNING;
+        this.title = 'Unsupported Request';
+    }
+
+    private get _message() {
+        switch (this.request.kind) {
+            case RequestType.SIGN_MESSAGE:
+                return 'Message signing is not supported for Ledgers.';
```
sisou:
```suggestion
                return 'Message signing is not yet supported for Ledgers.';
```

src/views/ErrorHandlerUnsupportedLedger.vue
```diff
@@ -0,0 +1,59 @@
...
+            case RequestType.EXPORT:
+                return 'Can not export a backup from a Ledger. Please refer to your 24 words backup you created '
+                    + 'when setting up your device.';
+            case RequestType.CHANGE_PASSWORD:
+                return 'To change the pin of your Ledger please go to the settings on your device.';
```
sisou:
```suggestion
                return 'To change the PIN of your Ledger please go to the settings on your device.';
```

src/router.ts
danimoh: Because for export and change password, the call should always reject.
For sign message, the user is given the option to chose another account unless the `forceSender` flag (not implemented yet) is set in which case he would be forwarded to the error page.

PR #258, opened 2019-05-13
Update loading animation

sisou:


sisou:
I updated the PR to only include the animation changes and removed all logo changes.

PR #261, opened 2019-05-14
Only reject if the corresponding rpcSate has a source

nibhar:
in conjunction with nimiq/rpc#22

Updates @nimiq/rpc to v0.3.0 and @nimiq/keyguard-client to v0.6.0.

requires nimiq/keyguard-next#333

### DIFF
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -273,7 +273,13 @@ export default class RpcApi {
                 }
 
                 if (error.message === 'Request aborted') {
-                    if (window.opener) {
+                    /*
+                     * In case the window is an iFrame and the recovered state is the one with which the
+                     * iFrame was initialized (has a source), then reject it. The iFrame will be closed as a result.
+                     * If not there was another history entry in between, where a history.back() will navigate to,
+                     * not closing the iFrame in the process
+                     */
+                    if (this._staticStore.rpcState!.source && window.opener) {
                         this.reject(error);
                     } else {
                         window.history.back();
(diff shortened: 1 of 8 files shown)


src/lib/RpcApi.ts
```diff
@@ -267,7 +267,9 @@ export default class RpcApi {
                 }
 
                 if (error.message === 'Request aborted') {
-                    if (window.opener) {
+                    const rpcState = RpcState.fromJSON(state.rpcState);
```
sisou: The rpcState is also available as `this._staticStore.rpcState`, because the state was already recovered in line 259.

nibhar: You are right. Fixed it.

src/lib/RpcApi.ts
```diff
@@ -267,7 +267,13 @@ export default class RpcApi {
                 }
 
                 if (error.message === 'Request aborted') {
-                    if (window.opener) {
+                    /*
+                     * In case the window is an iFrame and the recovered state is the one with which the
+                     * iFrame was initialized (has a source), then reject it. The iFrame will be closed as a result.
+                     * If not there was another history entry in between, where a history.back() will navigate to,
+                     * not closing the iFrame in the process
```
danimoh: iframe or popup?

sisou: I also think you mean Popups :D

nibhar: Oh yes, I absolutely mean popup...

PR #262, opened 2019-05-14
Checkout onboarding

nibhar:
Onboarding while in Checkout can now return to Checkout.

### DIFF
--- a/src/views/OnboardingSelector.vue
+++ b/src/views/OnboardingSelector.vue
@@ -1,16 +1,16 @@
 <template>
     <NotEnoughCookieSpace v-if='notEnoughCookieSpace'/>
-    <div v-else class="container" :class="{isCheckoutOnboarding}">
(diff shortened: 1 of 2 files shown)


src/views/OnboardingSelector.vue
```diff
@@ -7,7 +7,7 @@
 
             <button v-if="!request.disableBack" class="global-close nq-button-s" @click="close">
                 <ArrowLeftSmallIcon/>
-                Back to {{request.appName}}
+                Back to {{isCheckoutOnboarding ? 'Checkout' : (isChooseAddressOnboarding ? 'Choose Address' : request.appName)}}
```
sisou: Now with two conditions, this should be moved to a separate computed prop, like `backButtonText` that does to computing in the class instead of in the template.

sisou:
I think for completeness, we also have an address selector in the `SIGN_MESSAGE` request, so users could also come from there :wink:

PR #263, opened 2019-05-14
Client: Derived types

### LINKED ISSUE #252: Client: Derive result type of handlers from RequestType
Improves developer experience if TypeScript is used.

Bettelstab:
Add a bunch of generics to make TypeScript derive return types and expected types for resolve from RequestType and RequestBehavior.
Resolves #252

### DIFF
--- a/client/RequestBehavior.ts
+++ b/client/RequestBehavior.ts
@@ -21,10 +21,6 @@ export abstract class RequestBehavior<B extends BehaviorType> {
     ): Promise<B extends BehaviorType.REDIRECT ? void : ResultByRequestType<R>> {
         throw new Error('Not implemented');
     }
-
-    public get type(): B {
-        return this._type;
-    }
 }
 
 export enum BehaviorType {
--- a/client/RequestBehavior.ts
+++ b/client/RequestBehavior.ts
@@ -8,9 +8,9 @@ export abstract class RequestBehavior<B extends BehaviorType> {
         return url.origin;
     }
 
-    private readonly _type: BehaviorType;
+    private readonly _type: B;
 
-    constructor(type: BehaviorType) {
+    constructor(type: B) {
         this._type = type;
     }
 
@@ -22,7 +22,7 @@ export abstract class RequestBehavior<B extends BehaviorType> {
         throw new Error('Not implemented');
     }
 
-    public get type() {
+    public get type(): B {
         return this._type;
     }
 }
--- a/client/RequestBehavior.ts
+++ b/client/RequestBehavior.ts
@@ -1,5 +1,6 @@
 import { RequestType } from '../src/lib/RequestTypes';
 import { PostMessageRpcClient, RedirectRpcClient } from '@nimiq/rpc';
+import { ResultByRequestType } from '@/lib/PublicRequestTypes';
 
 export class RequestBehavior {
     public static getAllowedOrigin(endpoint: string) {
@@ -13,7 +14,11 @@ export class RequestBehavior {
         this._type = type;
     }
 
-    public async request(endpoint: string, command: RequestType, args: any[]): Promise<any> {
+    public async request<T extends RequestType, B extends RequestBehavior>(
+        endpoint: string,
+        command: T,
+        args: any[],
+    ): Promise<B extends RedirectRequestBehavior ? void : ResultByRequestType<T>> {
         throw new Error('Not implemented');
     }
 
@@ -22,7 +27,7 @@ export class RequestBehavior {
     }
 }
 
-enum BehaviorType {
+export enum BehaviorType {
     REDIRECT,
     POPUP,
     IFRAME,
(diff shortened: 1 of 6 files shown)


client/AccountsClient.ts
```diff
@@ -44,14 +39,15 @@ export default class AccountsClient {
...
     private readonly _iframeBehavior: IFrameRequestBehavior;
     private readonly _redirectClient: RedirectRpcClient;
 
-    constructor(endpoint: string = AccountsClient.DEFAULT_ENDPOINT, defaultBehavior?: RequestBehavior) {
+    constructor(endpoint: string = AccountsClient.DEFAULT_ENDPOINT,
+                defaultBehavior?: RequestBehavior<DB>) {
         this._endpoint = endpoint;
         this._defaultBehavior = defaultBehavior || new PopupRequestBehavior(
-            `left=${window.innerWidth / 2 - 400},top=75,width=800,height=850,location=yes,dependent=yes`);
+            `left=${window.innerWidth / 2 - 400},top=75,width=800,height=850,location=yes,dependent=yes`) as any;
```
danimoh: Why not rather properly type the options in the `PopupRequestBehavior` as `string`?

client/AccountsClient.ts
```diff
@@ -62,95 +58,130 @@ export default class AccountsClient {
...
+        resolve: (result: ResultByRequestType<T>, state: any) => void,
+        reject?: (error: Error, state: any) => void,
     ) {
         this._redirectClient.onResponse(command,
             // State is always an object containing at least the __command property
-            (result: RpcResult, rpcId, state) => resolve(result, state),
-            (error: Error, rpcId, state) => reject && reject(error, state),
+            (result: ResultByRequestType<T>, rpcId, state) => resolve(result, state),
+            (error: Error, rpcId, state) => {
+                if (reject) {
```
danimoh: +1 for removing the `reject && ...`.
I'd rather use `if (!reject) return;` though.

client/AccountsClient.ts
```diff
@@ -62,95 +58,130 @@ export default class AccountsClient {
...
+                    reject(error, state);
+                }
+            },
         );
     }
 
-    public onboard(request: BasicRequest, requestBehavior = this._defaultBehavior): Promise<Account> {
+    public onboard<B extends BehaviorType = DB>(
+        request: BasicRequest,
+        requestBehavior: RequestBehavior<B> = this._defaultBehavior as any,
```
danimoh: Why is the `as any necessary`?

client/AccountsClient.ts
```diff
@@ -62,95 +58,130 @@ export default class AccountsClient {
...
+    ) {
```
danimoh: Hum, does it really improve the developer experience to remove the return type?
I guess it can be automatically determined now, but the return type is now less obvious by looking at the code.

client/AccountsClient.ts
```diff
@@ -62,95 +58,130 @@ export default class AccountsClient {
...
         return this._request(requestBehavior, RequestType.MIGRATE, [{ appName: 'Accounts Client' }]);
     }
 
     /**
      * Only accessible in iframe from Nimiq domains.
      */
-    public list(requestBehavior = this._iframeBehavior): Promise<Account[]> {
+    public list<B extends BehaviorType = DB>(
+        requestBehavior: RequestBehavior<B> = this._iframeBehavior as any,
```
danimoh: Shouldn't this ideally be restricted to `IFrameRequestBehavior`?

src/lib/PublicRequestTypes.ts
```diff
@@ -139,3 +140,12 @@ export type RpcResult = SignedTransaction
                       | Address
                       | SignedMessage
                       | ExportResult;
+
+export type ResultByRequestType<T> =
+    T extends RequestType.ONBOARD | RequestType.SIGNUP | RequestType.RENAME? Account :
```
danimoh: Missing space in front of `?`

client/AccountsClient.ts
Bettelstab: It's not the options I type as any, but the created object. It's strange I have to do it, but otherwise the compiler complains about something like `B (extends BehaviorType) not assignable to DB (extends BehaviorType)', yet we know that everything works fine...

client/AccountsClient.ts
Bettelstab: see above :(

client/AccountsClient.ts
Bettelstab: It was not correct before, because it's dependent on the RequestBehavior. Annotating the true conditional type probably doesn't improve dev UX a lot.

client/AccountsClient.ts
Bettelstab: Could be. I guess it's kind of a backup option for Apple devices...

client/RequestBehavior.ts
```diff
@@ -1,34 +1,39 @@
...
         const url = new URL(endpoint);
         return url.origin;
     }
 
-    private readonly _type: BehaviorType;
+    private readonly _type: B;
 
-    constructor(type: BehaviorType) {
+    constructor(type: B) {
         this._type = type;
```
nibhar: Is this used? I can't seem to find it anywhere.

Bettelstab: Turns out it's not used. But I can't get type magic to work if I delete the whole `BehaviorType`.

nibhar: Could then still be removed as it is unnecessary.

PR #272, opened 2019-05-15
Style update to reflect changes in keyguard in SignTransactionLedger

danimoh:
Adapting to Keyguard changes made in https://github.com/nimiq/keyguard-next/pull/309.

### DIFF
--- a/src/views/SignTransactionLedger.vue
+++ b/src/views/SignTransactionLedger.vue
@@ -24,13 +24,11 @@
 
             <hr class="blur-target">
 
-            <div class="value nq-light-blue blur-target">
-                {{ transactionValue }}
-                <span class="nim-symbol">NIM</span>
-            </div>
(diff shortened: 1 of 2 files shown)


src/views/SignTransactionLedger.vue
```diff
@@ -378,9 +371,9 @@ export default class SignTransactionLedger extends Vue {
     }
 
     .data {
-        margin: .5rem 7rem 0;
+        margin: .25rem 3rem 0;
         opacity: 1;
-        text-align: center;
+        color: var(--nimiq-blue);
```
sisou: Shouldn't the default body color be nimiq-blue? Why is it necessary to specify it here?

danimoh: `nq-text` has `rgba(31,35,72,.7);` set as color.

PR #274, opened 2019-05-16
Show loader while router loads route

sisou:
Following advice on https://stackoverflow.com/a/44887358/4204380 and the docs on https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards, this makes the loading spinner display while the router is (lazy-)loading a route.

This prevents the loading spinner from disappearing as soon as the request/result is known but before the route's component is shown (because the component loading is only started after the request/result is set) and thus presents a "gap in loading".

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -40,18 +40,20 @@ if (window.location.origin === 'https://hub.nimiq-testnet.com') {
 }
 
 const app = new Vue({
-  data: { loading: false },
+  data: { loading: true },
   router,
   store,
   render: (h) => h(App),
 }).$mount('#app');
 
+let _loadingTimeout: number = 0;
 router.beforeEach((to, from, next) => {
-  app.loading = true;
+   // Only show loader when lazy-loading takes longer than 500ms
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -39,12 +39,22 @@ if (window.location.origin === 'https://hub.nimiq-testnet.com') {
...
-new Vue({
+const app = new Vue({
+  data: { loading: false },
   router,
   store,
   render: (h) => h(App),
 }).$mount('#app');
 
+router.beforeEach((to, from, next) => {
+  app.loading = true;
```
danimoh: On a fast internet connecting, this causes flickering.
I'd only set this after a short delay.

src/main.ts
```diff
@@ -39,12 +39,22 @@ if (window.location.origin === 'https://hub.nimiq-testnet.com') {
   });
 }
 
-new Vue({
+const app = new Vue({
+  data: { loading: false },
```
danimoh: Does it have to be `false`?
`true` would make more sense to the reader.

PR #278, opened 2019-05-16
Change walletId to be a 32 byte hash in the LedgerApi

danimoh:


src/lib/LedgerApi.ts
```diff
@@ -58,7 +58,7 @@
 // - ledger already handling another request (from another tab)
 
 // tslint:disable-next-line:max-line-length
-// TODO: change implementation to be flow typed, integrate into ledger-api repository and bundle with ledger provided flow libraries directly. LedgerJs:any type is just a temporary workaround
+// TODO: change implementation to be flow typed, integrate into ledger-api repository and bundle with ledger provided flow libraries directly. LedgerJs:any type is just a temporary workaround. Also, the verification and address computation in ledgerjs should be done by Nimiq's crypto methods instead of unnecessarily bundling tweetnacl and blakejs.
```
sisou: Please make this a multi-line comment, favourably line-broken at ~120 chars.

src/lib/WalletInfoCollector.ts
```diff
@@ -125,9 +125,15 @@ export default class WalletInfoCollector {
...
+        // For ledger retrieve the keyId from the connected Ledger.
+        // Doing this after starting deriveAccounts, as the call is cheaper then.
+        if (walletType === WalletType.LEDGER) {
+            keyId = await LedgerApi.getWalletId();
+        }
+
         // Get or create the walletInfo instance and derive the first set of derived accounts
         const [walletInfo, firstSetOfDerivedAccounts] = await Promise.all([
-            WalletInfoCollector._getWalletInfoInstance(walletType, keyId),
+            WalletInfoCollector._getWalletInfoInstance(walletType, keyId!),
```
sisou: Instead of using the `!notnull` operator here, I would prefer if the parameter in the function would have a default value instead of being nullable.

That way there is really no way that the keyId is empty, even when this method is used with the keyId omitted.

src/lib/WalletInfoCollector.ts
```diff
@@ -223,7 +229,7 @@ export default class WalletInfoCollector {
         }
 
         const releaseKey = walletType === WalletType.BIP39
-            ? (removeKey?: boolean) => WalletInfoCollector._keyguardClient!.releaseKey(keyId, removeKey)
+            ? (removeKey?: boolean) => WalletInfoCollector._keyguardClient!.releaseKey(keyId!, removeKey)
```
sisou: Same for here.

src/lib/WalletInfoCollector.ts
```diff
@@ -249,13 +255,7 @@ export default class WalletInfoCollector {
...
 
     private static async _getWalletInfoInstance(walletType: WalletType, keyId: string): Promise<WalletInfo> {
-        let walletId: string;
-        if (walletType === WalletType.LEDGER) {
-            // for Ledger, we retrieve the walletId from the currently connected ledger
-            walletId = await LedgerApi.getWalletId();
-        } else {
-            walletId = await WalletStore.deriveId(keyId);
-        }
+        const walletId = await WalletStore.deriveId(keyId);
```
sisou: :+1: for less WalletType-conditions.

PR #288, opened 2019-05-20
Add salt to wallet id hash for ledger accounts

danimoh:
Proposal for salting the public key before the wallet id computation.
Not entirely sure whether we get a lot of upside from doing so, but at the same time it also doesn't hurt.

### DIFF
--- a/src/lib/LedgerApi.ts
+++ b/src/lib/LedgerApi.ts
@@ -385,6 +385,7 @@ class LedgerApi {
     }
 
     // private fields and methods
+    private static readonly WALLET_ID_SALT = 42;
     private static _apiPromise: Promise<LedgerJs> | null = null;
     private static _currentState: LedgerApi.State = { type: 'idle' as LedgerApi.StateType };
     private static _currentRequest: LedgerApiRequest<any> | null = null;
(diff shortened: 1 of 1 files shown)


src/lib/LedgerApi.ts
```diff
@@ -385,6 +385,7 @@ class LedgerApi {
     }
 
     // private fields and methods
+    private static readonly WALLET_ID_SALT = 42;
```
sisou: A fixed salt still allows a pre-computed rainbow attack.

danimoh: Rainbow tables are not really feasible here. The hash function is still fed 32 bytes of random data.
But no one would use a rainbow table here anyways. What's more of a problem is, that the wallet ID is easy to calculate if the public key is known.
What are you suggesting? To create a random salt and store that in a database in the hub along with the hashes? Do you _really_ think that's worth the effort?

sisou: No, I think hashing twice is better.

nibhar: Double hashing does not prevent the walletId being computable once the public key is known. A random salt stored in the database does. So if we decide that it should not be computable, then a random salt should be used in my opinion.

danimoh:
Closing this in favor of https://github.com/nimiq/hub/pull/298

PR #296, opened 2019-05-21
Rename Loader to StatusScreen

Bettelstab:
Resolves https://github.com/nimiq/hub/issues/142

### DIFF
--- a/src/components/StatusScreen.vue
+++ b/src/components/StatusScreen.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="loader" :class="{
+    <div class="status-screen" :class="{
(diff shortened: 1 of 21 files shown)


src/components/StatusScreen.vue
```diff
@@ -279,7 +279,7 @@ export default Loader;
         position: relative;
     }
 
-    .loader.small .status-row {
+    .status-screen .status-row {
```
danimoh: Here you removed the `.small` part.
On purpose?

sisou:
Using the word "screen" here is the first use of this word in the whole codebase. Everything else is called a "page". Just saying.

PR #297, opened 2019-05-21
Whitelist Public API methods

### LINKED ISSUE #294: Whitelist only certain methods for use by 3rd parties
Only allow access from 3rd-parties (which are not the Safe or the Hub itself) to

- `checkout()`
- `signTransaction()`
- `signMessage()`
- `chooseAddress()`

Access to account managing methods will then be enabled when a proper authentication/permission scheme is in place.

sisou:
And block access to account management methods from non-privileged origins.

Resolves #294.


README.md
```diff
@@ -14,14 +14,6 @@ interact with their users' Nimiq addresses.
...
         - [Choose Address](#choose-address)
         - [Sign Transaction](#sign-transaction)
-        - [Signup](#signup)
-        - [Login](#login)
-        - [Onboard](#onboard)
-        - [Logout](#logout)
-        - [Export](#export)
-        - [Change Password](#change-password)
-        - [Add Address](#add-address)
-        - [Rename](#rename)
```
nibhar: I wouldn't remove them from the documentation. State that they are not (yet) publicly available but restricted to the Safe instead. Locally developers can already build against those endpoints.

styppo: Makes sense to me as well :+1: Maybe put them under a section "wallet management" or something in the documentation, stating that these endpoints will be opened up for third-party use in an upcoming version.

danimoh:
So, was the documentation now deleted, not moved?

PR #301, opened 2019-05-23
Preserve a disableBack flag through the parser.

sisou:
I am not sure if we should add the flag to the `BasicRequest`, as that would mean it would be available for every request, which it isn't. It's only respected in the onboarding menu.


sisou:
The bug that this fixes:
The Safe sends the onboading request (when no accounts exist) with the `disabledBack` flag, and the `OnboardingSelector` view also respects this setting. 

But the flag never arrives at the view, because the parser does not forward it.

PR #310, opened 2019-06-11
Replace outdated vue-raven.js by official Sentry SDK

sisou:


sisou:
This increases the vendor-chunk from 2.2 MB to 2.46 MB (gzipped: from 575.61 KB to 635.97 KB) for the testnet (mainnet is 36 KB less because client/rpc is not included (included in testnet for the demo page)).

I think this is an acceptable trade-off, wdyt?

src/main.ts
```diff
@@ -41,10 +42,12 @@ const rpcApi = new RpcApi(store, staticStore, router);
...
 
 if (Config.reportToSentry) {
-    Vue.use(VueRaven, {
+    initSentry({
         dsn: 'https://92f2289fc2ac4c809dfa685911f865c2@sentry.io/1330855',
+        integrations: [new SentryVueIntegration({Vue, attachProps: true})],
         environment: Config.network,
     });
+    Vue.prototype.$captureException = captureException;
 }
```
danimoh: The Sentry setup could be moved to `Helpers.ts` to avoid code duplciation.

danimoh:
The current version is 5.11.1 so would make sense to already update to that.

PR #315, opened 2019-07-02
Fix wording in AddAddress IdenticonSelector, autofocus LabelInput

sisou:


src/lib/Constants.ts
```diff
@@ -39,3 +39,11 @@ export const ERROR_CANCELED = 'CANCELED';
 export const ERROR_INVALID_NETWORK = 'Invalid network name';
 export const ERROR_TRANSACTION_RECEIPTS = 'Failed to retrieve transaction receipts for';
 export const ERROR_COOKIE_SPACE = 'Not enough cookie space';
+
+// Input
+export const MIN_WIDTH_FOR_AUTOFOCUS = 600; // px
+export function getDocumentWidth() {
```
nibhar: As these two only ever get used in that specific context it could also just be 
```typescript
export function autofocusEnabled() {
  return (window.innerWidth 
    || document.documentElement.clientWidth
    || document.body.clientWidth ) < 600;
}
```
If necessary or deemed beneficial the constant `MIN_WIDTH_FOR_AUTOFOCUS` could also stay. I do not think we use it otherwise yet.

sisou: Then I should really rename the method to `isDesktop()`, because that what it actually checks for.

PR #317, opened 2019-07-24
Show error page instead of redirect in case of missing request

sisou:
This fixes the issue of a redirect happening in a popup, when the popup is initially blocked by a popup-blocker but then opened anyway from the browser prompt. In those cases, the reference to the popup is lost in the opening window and the popup receives no postMessage request within the timeout. Additionally, because the reference from the opening window got lost, the `window.opener` in the popup is now `null`, triggering the redirect case, opening the Safe in the popup.

I believe we don't need a redirect-on-error and can just display the error page. The error page then detects if the user can go back to the sending page, or close the window otherwise.

### DIFF
--- a/src/config/config.testnet.ts
+++ b/src/config/config.testnet.ts
@@ -8,6 +8,5 @@ export default {
         'https://safe.nimiq-testnet.com',
         'https://hub.nimiq-testnet.com', // For testing with the deployed demos.html page
     ],
-    redirectTarget: 'https://safe.nimiq-testnet.com',
     reportToSentry: true,
 };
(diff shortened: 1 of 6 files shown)


src/config/config.testnet.ts
```diff
@@ -8,6 +8,5 @@ export default {
         'https://safe.nimiq-testnet.com',
         'https://hub.nimiq-testnet.com', // For testing with the deployed demos.html page
     ],
-    redirectTarget: 'https://safe.nimiq-testnet.com',
```
danimoh: `redirectTarget` should be removed from the readme too.

PR #320, opened 2019-07-26
Add new accounts to Vuex when they are created, not when they are needed

sisou:
To avoid having to actively check for and add newly added accounts from a side-request, the side-requests should add their created/logged-in accounts to the Vuex store automatically. This reduces code and headaches.

### DIFF
--- a/src/views/ChooseAddress.vue
+++ b/src/views/ChooseAddress.vue
@@ -48,8 +48,6 @@ export default class ChooseAddress extends Vue {
     }) => any;
 
     public async created() {
-        await this.handleOnboardingResult();
-
         if (this.processedWallets.length === 0) {
             this.goToOnboarding(true);
         }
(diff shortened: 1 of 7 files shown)


src/views/ChooseAddress.vue
```diff
@@ -87,31 +85,6 @@ export default class ChooseAddress extends Vue {
...
-    private async handleOnboardingResult() {
-        // Check if we are returning from an onboarding request
-        if (staticStore.sideResult && !(staticStore.sideResult instanceof Error)) {
-            const sideResult = staticStore.sideResult as Account[];
-
-            // Add imported wallets to Vuex store
-            for (const account of sideResult) {
-                const walletInfo = await WalletStore.Instance.get(account.accountId);
-                if (walletInfo) {
-                    this.$addWallet(walletInfo);
```
danimoh: `$addWallet` does now not need to be imported anymore.

src/views/LoginSuccess.vue
```diff
@@ -35,6 +35,8 @@ export default class LoginSuccess extends Vue {
     @Static private request!: ParsedBasicRequest;
     @State private keyguardResult!: KeyguardClient.KeyResult;
 
+    @Action('addWalletAndSetActive') private $addWalletAndSetActive!: (walletInfo: WalletInfo) => any;
```
danimoh: Or `=> void`?

src/views/SignMessage.vue
```diff
@@ -130,30 +128,6 @@ export default class SignMessage extends Vue {
...
-    private async handleOnboardingResult() {
-        // Check if we are returning from an onboarding request
-        if (staticStore.sideResult && !(staticStore.sideResult instanceof Error)) {
-            const sideResult = staticStore.sideResult as Account[];
-
-            // Add imported wallets to Vuex store
-            for (const account of sideResult) {
-                const walletInfo = await WalletStore.Instance.get(account.accountId);
-                if (walletInfo) {
-                    this.$addWallet(walletInfo);
```
danimoh: `$addWallet` does not need to be imported anymore.

src/views/LoginSuccess.vue
sisou: Generally yes, but all other occurances of the 'addWallet' and 'setActiveAccount' actions are also `=> any`, so I kept with tradition.

PR #321, opened 2019-07-26
Warn about potentially incomplete addresses on ledger signup

danimoh:
Warn user that addresses are potentially incomplete if transaction receipts could not be retrieved successfully.
This behavior is similar to the incomplete import of Keyguard accounts (https://github.com/nimiq/hub/pull/179).
However, the Ledger flow eventually ends in a success screen, even if there was a warning which I consider more consistent. (Edit: I just saw LoginSuccess had been changed to this behavior in a later commit too)

### DIFF
--- a/src/views/SignupLedger.vue
+++ b/src/views/SignupLedger.vue
@@ -8,8 +8,10 @@
             <transition name="transition-fade">
                 <StatusScreen v-if="state === constructor.State.LOADING
                         || state === constructor.State.FETCHING_ADDRESSES
+                        || state === constructor.State.FETCHING_INCOMPLETE
                         || state === constructor.State.FINISHED"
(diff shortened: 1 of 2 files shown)


src/views/SignupLedger.vue
```diff
@@ -104,23 +117,34 @@ export default class SignupLedger extends Vue {
         }
     }
 
+    private get statusScreenMessage() {
+        if (this.state !== SignupLedger.State.FETCHING_INCOMPLETE) return '';
+        else return 'We might have missed used addresses that have no balance.';
```
nibhar: I can't recall ever using "we" in user facing text. I suggest replacing it i.e.
```suggestion
        else return 'Used addresses without balance might have been missed.';
```

PR #326, opened 2019-09-15
Ledger atomic requests

danimoh:
See commit history.

### DIFF
--- a/src/lib/LedgerApi.ts
+++ b/src/lib/LedgerApi.ts
@@ -496,9 +496,7 @@ class LedgerApi {
                 LedgerApi._throwError(LedgerApi.ErrorType.WRONG_LEDGER, e);
             }
             LedgerApi._currentlyConnectedWalletId = null;
-            if (message.indexOf('dependencies') !== -1) {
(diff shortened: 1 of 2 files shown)


src/lib/LedgerApi.ts
```diff
@@ -284,7 +285,14 @@ class LedgerApi {
         const request = new LedgerApiRequest(LedgerApi.RequestType.CONFIRM_ADDRESS,
             async (api, params): Promise<string> => {
                 const result = await api.getAddress(params.keyPath, /*validate*/ true, /*display*/ true);
-                return result.address;
+                const confirmedAddress = result.address;
+
+                if (params.addressToConfirm!.replace(/ /g, '').toUpperCase()
+                    !== confirmedAddress.replace(/ /g, '').toUpperCase()) {
+                    LedgerApi._throwError(LedgerApi.ErrorType.REQUEST_ASSERTION_FAILED, 'Address mismatch', request);
+                }
```
sisou: Does `LedgerApi._throwError` actually `throw` an `Error`? Or do you need to `return` here, to avoid resolving this request?

danimoh: Yes, it actually throws.

PR #327, opened 2019-09-15
Allow requests to be Promises.

nibhar:


client/HubApi.ts
```diff
@@ -84,28 +84,28 @@ export default class HubApi<DB extends BehaviorType = BehaviorType.POPUP> { // D
      */
 
     public checkout<B extends BehaviorType = DB>(
-        request: CheckoutRequest,
+        request: Promise<CheckoutRequest>,
```
sisou: When the type is only promise now, is it still OK and not a type error when a regular object is passed?

demos/Demo.ts
```diff
@@ -84,7 +84,11 @@ class Demo {
         document.querySelector('button#sign-transaction-popup').addEventListener('click', async () => {
             const txRequest = generateSignTransactionRequest();
             try {
-                const result = await demo.client.signTransaction(txRequest);
+                const result = await demo.client.signTransaction(
+                    new Promise<SignTransactionRequest>((resolve, reject) => {
+                        window.setTimeout(() => resolve(txRequest), 5000);
```
sisou: 5s is a bit long, how about 2s?

client/HubApi.ts
nibhar: yes

demos/Demo.ts
nibhar: It is a demo... I can change it to 2s though.

client/HubApi.ts
danimoh: Nope, that's not true. `CheckoutRequest` objects can't be passed instead of `Promise<CheckoutRequest>` as `CheckoutRequest` doesn't have `then` or `catch`.

client/RequestBehavior.ts
```diff
@@ -17,7 +17,7 @@ export abstract class RequestBehavior<B extends BehaviorType> {
     public async request<R extends RequestType>(
         endpoint: string,
         command: R,
-        args: any[],
+        args: Array<Promise<any>>,
```
danimoh: In `HubApi._request` the type of `args` that is passed to this request method is `Array<Promise<any>> | any[]`. Why the inconsistency?
I'd recommend `Iterable<any | PromiseLike<any>>` as type for `args` in `_request` and `request` as this seems to be the type that `Promise.all` accepts.

PR #341, opened 2019-11-04
Create API handler functions only once

sisou:
This PR changes the creation of the Hub and Keyguard API handler functions to only be created once and then re-used, instead of creating huge methods multiple times, once for each API method.

This should slightly reduce initialization time for the Hub and reduce memory usage.

The diff should be viewed with whitespace ignored, because the indentation of big blocks of code changed.

### DIFF
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -17,11 +17,7 @@ import {
     KeyguardCommand,
     Errors,
     ObjectType,
-    SignatureResult,
-    DerivedAddress,
-    SingleKeyResult,
-    ExportResult,
-    SimpleResult,
+    ResultByCommand,
 } from '@nimiq/keyguard-client';
 import { keyguardResponseRouter, REQUEST_ERROR } from '@/router';
 import { StaticStore } from '@/lib/StaticStore';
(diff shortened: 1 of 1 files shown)


src/lib/RpcApi.ts
```diff
@@ -22,6 +32,8 @@ import { Raven } from 'vue-raven'; // Sentry.io SDK
 import { ERROR_CANCELED } from './Constants';
 import Config from 'config';
 
+type KeyguardResult = SignatureResult | DerivedAddress[] | SingleKeyResult[] | ExportResult | SimpleResult;
```
sisou: Sadly there is no combined type export from the Keyguard client yet, so all result types have to be manually combined here.

nibhar: could also be
```suggestion
type KeyguardResult = RedirectResult | ExportResult;
```

`RedirectResult` is supposed to be the combined type but is missing `ExportResult`. I created nimiq/keyguard/pull/383 to include `ExportResult` in `RedirectResult`.

src/lib/RpcApi.ts
```diff
@@ -276,52 +290,59 @@ export default class RpcApi {
...
+            this._keyguardClient.on(command,
+                (result, state) => this._keyguardSuccessHandler(command, result, state),
+                (error, state) => this._keyguardErrorHandler(command, error, state),
+            );
+        }
+    }
 
-                // Set result
-                this._store.commit('setKeyguardResult', result);
+    private _keyguardSuccessHandler(command: KeyguardCommand, result: KeyguardResult, state?: ObjectType | null) {
```
nibhar: Alternatively I think this could also work and would make a combined type unnecessary here:

```suggestion
    private _keyguardSuccessHandler<C extends KeyguardCommand>(command: C, result: ResultByCommand<C>, state?: ObjectType | null) {
```

sisou: This! I love this! :heart: :+1:

PR #342, opened 2019-11-14
Fix check for sufficient iOS cookie size on Login

danimoh:
Note that I didn't test this fix.

### DIFF
--- a/src/views/OnboardingSelector.vue
+++ b/src/views/OnboardingSelector.vue
@@ -34,10 +34,8 @@ export default class OnboardingSelector extends Vue {
     private notEnoughCookieSpace = false;
 
     public async created() {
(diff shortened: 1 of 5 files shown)


src/views/OnboardingSelector.vue
```diff
@@ -34,10 +34,8 @@ export default class OnboardingSelector extends Vue {
     private notEnoughCookieSpace = false;
 
     public async created() {
-        if ((BrowserDetection.isIOS() || BrowserDetection.isSafari()) && !await CookieHelper.canFitNewWallets()) {
-            this.notEnoughCookieSpace = true;
-            return;
-        }
+        this.notEnoughCookieSpace = (BrowserDetection.isIOS() || BrowserDetection.isSafari())
+            && !await CookieHelper.canFitNewWallets();
     }
```
sisou: It looks like the `return;` statement got lost here. Is it not needed anymore?

PR #347, opened 2019-11-25
Allow external recipient label for sign-transaction requests

sisou:
Forward recipient label to Keyguard to display there.

This was previously part of the `sebastian/cashlink-create-manage` branch.


sisou:
To display contact labels all the way in the Keyguard, instead of the cut-off address as the identicon label.

PR #351, opened 2019-11-26
Move extraData from MultiCurrencyCheckoutRequest.extraData to NimiqDirectPaymentOptions.protocolSpecific.extraData.

nibhar:
This PR moves the extraData from `MultiCurrencyCheckout.extraData` to `NimiqDirectPaymntOptions.protoclSpecific.extraData` as it is only ever used in that case.

Additionally `fee`, `feePerByte` and `extraData` can now be `undefined` in `NimiqSpecifics` as `ParsedPaymentOptions.update()` determines the need for updating a value based on wether it is `undefined` or not. The constructor of `ParsedPaymentOptions` got a new argument `enforceDefaultValues` to account for the change.

### DIFF
--- a/src/lib/RequestParser.ts
+++ b/src/lib/RequestParser.ts
@@ -88,14 +88,6 @@ export class RequestParser {
                     }
                 }
 
-                if (checkoutRequest.extraData !== undefined && typeof checkoutRequest.extraData !== 'string'
-                    && !(checkoutRequest.extraData instanceof Uint8Array)) {
-                    throw new Error('extraData must be a string or Uint8Array');
-                }
-                const data = typeof checkoutRequest.extraData === 'string'
-                    ? Utf8Tools.stringToUtf8ByteArray(checkoutRequest.extraData)
-                    : checkoutRequest.extraData || new Uint8Array(0);
-
                 if (!checkoutRequest.version || checkoutRequest.version === 1) {
                     if (typeof checkoutRequest.value !== 'number' || checkoutRequest.value <= 0) {
                         throw new Error('value must be a number >0');
@@ -106,7 +98,6 @@ export class RequestParser {
                         version: 1,
                         appName: checkoutRequest.appName,
                         shopLogoUrl: checkoutRequest.shopLogoUrl,
-                        data,
                         time: Date.now(),
                         paymentOptions: [new ParsedNimiqDirectPaymentOptions({
                             currency: Currency.NIM,
@@ -114,6 +105,7 @@ export class RequestParser {
                             amount: checkoutRequest.value.toString(),
                             expires: 0, // unused for NimiqCheckoutRequests
                             protocolSpecific: {
+                                extraData: checkoutRequest.extraData,
                                 recipient: checkoutRequest.recipient,
                                 recipientType: checkoutRequest.recipientType || Nimiq.Account.Type.BASIC,
                                 sender: checkoutRequest.sender,
@@ -122,7 +114,7 @@ export class RequestParser {
                                 flags: checkoutRequest.flags || Nimiq.Transaction.Flag.NONE,
                                 validityDuration: checkoutRequest.validityDuration,
                             },
-                        }, data)],
+                        })],
                     } as ParsedCheckoutRequest;
                 }
 
@@ -196,7 +188,6 @@ export class RequestParser {
                         shopLogoUrl: checkoutRequest.shopLogoUrl,
                         callbackUrl: checkoutRequest.callbackUrl,
                         csrf: checkoutRequest.csrf,
-                        data,
                         time: !checkoutRequest.time
                             ? Date.now()
                             : isMilliseconds(checkoutRequest.time)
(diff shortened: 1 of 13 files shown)


src/lib/RequestParser.ts
```diff
@@ -214,7 +205,17 @@ export class RequestParser {
...
+                                            // Once extraData from MultiCurrencyCheckoutRequest is removed
+                                            // the next few lines become obsolete.
+                                            if (!option.protocolSpecific.extraData) {
+                                                option.protocolSpecific.extraData = checkoutRequest.extraData;
+                                                if (option.protocolSpecific.extraData) {
+                                                    console.warn('Usage of MultiCurrencyCheckoutRequest.extraData is'
+                                                        + ' deprecated. Use NimiqDirectPaymentOptions.protocolSpecific'
+                                                        + '.extraData instead');
+                                                }
+                                            }
```
sisou: Suggestion to put the warning an indentation level higher:

```js
if (checkoutRequest.extraData) {
    console.warn(...);
}
if (!option.protocolSpecific.extraData) {
    option.protocolSpecific.extraData = checkoutRequest.extraData;
}
```
We could even move the assignment into that first condition, which would always overwrite any data in protocolSpecific with the global extraData (which I think is reasonable).

src/lib/paymentOptions/NimiqPaymentOptions.ts
```diff
@@ -14,23 +15,31 @@ export interface NimiqSpecifics {
...
 export class ParsedNimiqDirectPaymentOptions extends ParsedPaymentOptions<Currency.NIM, PaymentType.DIRECT> {
     public amount: number;
-    private extraData: Uint8Array;
 
-    public constructor(options: NimiqDirectPaymentOptions, extraData: Uint8Array) {
+    public constructor(options: NimiqDirectPaymentOptions, enforceDefaultValues = true) {
         super(options);
-        this.extraData = extraData;
+
+        if (options.protocolSpecific.extraData !== undefined && typeof options.protocolSpecific.extraData !== 'string'
```
sisou: I think the first check can be simply a truthy check. Or do you want to allow an empty string?

src/lib/paymentOptions/NimiqPaymentOptions.ts
```diff
@@ -81,24 +90,29 @@ export class ParsedNimiqDirectPaymentOptions extends ParsedPaymentOptions<Curren
...
+            // Note that the transaction size can be bigger than this, for example if the sender type the user wants
+            // to use requires an extended transaction or if an extended transaction includes a multi signature proof.
+            // The size is therefore just an estimate. In the majority of cases the estimate will be accurate though
+            // and a fee that is slightly off will generally not be a problem.
+            const estimatedTransactionSize = requiresExtendedTransaction
+                ? 166 + (extraData ? extraData.byteLength : 0)
+                : 138;
+
+            if (fee === undefined) {
+                // xxx
```
sisou: What does this comment mean?

src/views/Checkout.vue
```diff
@@ -89,16 +89,16 @@ import { ERROR_CANCELED } from '../lib/Constants';
...
 
     @Static private rpcState!: RpcState;
     @Static private request!: ParsedCheckoutRequest;
-    private choosenCurrency: Currency | null = null;
-    private selectedCurrency: Currency = Currency.NIM;
-    private leftSlide!: Currency;
-    private rightSlide!: Currency;
-    private availableCurrencies: Currency[] = [];
+    private choosenCurrency: PublicCurrency | null = null;
+    private selectedCurrency: PublicCurrency = Checkout.Currency.NIM;
```
sisou:
```suggestion
    private selectedCurrency: PublicCurrency = PublicCurrency.NIM;
```

src/lib/RequestParser.ts
nibhar: I certainly don't think that the deprecated version should have priority over the new version.

The check could move up, but works just as well as it is.

src/lib/paymentOptions/NimiqPaymentOptions.ts
nibhar: It should not throw for an empty string. The prior default was `new Uint8Array(0)`, which kind of is similar to `''`.

src/lib/paymentOptions/NimiqPaymentOptions.ts
nibhar: It means it should be removed 👍

src/lib/paymentOptions/NimiqPaymentOptions.ts
```diff
@@ -81,24 +90,29 @@ export class ParsedNimiqDirectPaymentOptions extends ParsedPaymentOptions<Curren
...
+                if (feePerByte === undefined) {
+                    feePerByte = 0;
+                    fee = 0;
+                } else {
+                    fee = Math.ceil(feePerByte * estimatedTransactionSize);
+                }
             } else {
-                fee = Math.ceil(feePerByte * estimatedTransactionSize);
+                feePerByte = fee / estimatedTransactionSize;
             }
```
danimoh: I think this will now not update the `fee` correctly if only `feePerByte` is provided in an update (and vice versa).

src/views/Checkout.vue
```diff
@@ -155,13 +155,13 @@ export default class Checkout extends Vue {
...
     }
+}
 
-    private data() {
-        return {
-            Currency,
-        };
-    }
+namespace Checkout {
+    export const Currency = PublicCurrency;
```
danimoh: I have a slight preference for adding `export type Currency = PublicCurrency` here and using `Checkout.Currency` instead of `PublicCurrency` in the type annotations.
But if you prefer the way you did it, just keep it that way.

src/views/SignTransactionLedger.vue
```diff
@@ -385,27 +384,30 @@ export default class SignTransactionLedger extends Vue {
...
+        let data;
         let flags;
         if (this.request.kind === RequestType.SIGN_TRANSACTION) {
             const signTransactionRequest = this.request as ParsedSignTransactionRequest;
+            data = signTransactionRequest.data;
             flags = signTransactionRequest.flags;
         } else {
+            data = this.checkoutPaymentOptions!.protocolSpecific.extraData;
             flags = this.checkoutPaymentOptions!.protocolSpecific.flags;
         }
```
danimoh: Using destructuring here would be nice as is done for other assignments in this class.

PR #352, opened 2019-11-28
Add defer attributes to script by webpack plugin

sisou:
This makes it possible to also add the defer attribut to the Nimiq lib script, because the plugin also works in dev mode (the `defer.sh` bash script did not).


sisou:
@danimoh I force pushed a fix: The plugin is now loaded with `chainWebpack`, which adds it _after_ Vue's own plugins, thereby running after the script tags have been created and not failing with an undefined target anymore.

PR #354, opened 2019-11-30
HubApi Popup Overlay

sisou:
Adds an overlay when the Hub is opened in a popup. Clicking on this overlay refocuses the popup.

Tested on Chrome (Desktop and Android), Firefox (Desktop), Safari (Mac). More tests are welcome!

I added some helper functions and variables so that the code can be better minified for the standalone builds (DOM variables & methods can't be minified, so abstracting them is the way to go).

Paypal Overlay for reference (click the "Try the buttons" buttons): https://developer.paypal.com/docs/checkout/

image

### DIFF
--- a/client/RequestBehavior.ts
+++ b/client/RequestBehavior.ts
@@ -140,9 +140,11 @@ export class PopupRequestBehavior extends RequestBehavior<BehaviorType.POPUP> {
         // Explainer text
         const text = createElement('div');
         const textStyle = text.style;
-        appendChild(text, createTextNode('A popup is active,'));
+        appendChild(text, createTextNode('A popup has been opened,'));
         appendChild(text, createElement('br'));
-        appendChild(text, createTextNode('click anywhere to refocus.'));
+        appendChild(text, createTextNode('click anywhere to bring it back to the front.'));
+        // tslint:disable-next-line max-line-length
+        textStyle.fontFamily = 'Muli, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif';
         textStyle.fontSize = '24px';
         textStyle.fontWeight = '600';
         textStyle.lineHeight = '40px';
--- a/client/RequestBehavior.ts
+++ b/client/RequestBehavior.ts
@@ -81,6 +81,10 @@ export class PopupRequestBehavior extends RequestBehavior<BehaviorType.POPUP> {
         const origin = RequestBehavior.getAllowedOrigin(endpoint);
 
         const popup = this.createPopup(endpoint);
+
+        // Add page overlay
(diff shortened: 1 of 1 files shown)


client/RequestBehavior.ts
```diff
@@ -101,6 +108,81 @@ export class PopupRequestBehavior extends RequestBehavior<BehaviorType.POPUP> {
...
+        // can be animated, but also because it actually feels better when there is a short
+        // delay between the opening popup and the background fading.
+        setTimeout(() => overlay.style.opacity = '1', 100);
+
+        return appendChild(document.body, overlay) as HTMLDivElement;
+    }
+
+    private removeOverlay(overlay: HTMLDivElement): void {
+        overlay.style.opacity = '0';
+        setTimeout(() => document.body.removeChild(overlay), 400);
```
nibhar: Easing function on opacity runs for `.6`, why `400` here?

sisou:
Because it will be used on all websites that use the HubApi, not only Nimiq apps, so we have no control over the `rem` size.

client/RequestBehavior.ts
sisou: So the website becomes interactive earlier, when the overlay is barely visible anymore anyway.

client/RequestBehavior.ts
```diff
@@ -101,6 +108,81 @@ export class PopupRequestBehavior extends RequestBehavior<BehaviorType.POPUP> {
...
+        overlayStyle.transition = 'opacity 0.6s ease';
+        overlay.addEventListener('click', () => popup.focus());
+
+        // Top flex spacer
+        appendChild(overlay, createElement('div'));
+
+        // Explainer text
+        const text = createElement('div');
+        const textStyle = text.style;
+        appendChild(text, createTextNode('A popup is active,'));
```
danimoh: Sounds weird to me.
How about 'A popup has been opened'?

client/RequestBehavior.ts
```diff
@@ -101,6 +108,81 @@ export class PopupRequestBehavior extends RequestBehavior<BehaviorType.POPUP> {
...
+        appendChild(text, createElement('br'));
+        appendChild(text, createTextNode('click anywhere to refocus.'));
```
danimoh: 'click anywhere to bring it back to the front.'?

client/RequestBehavior.ts
```diff
@@ -101,6 +108,81 @@ export class PopupRequestBehavior extends RequestBehavior<BehaviorType.POPUP> {
...
+        textStyle.fontSize = '24px';
+        textStyle.fontWeight = '600';
+        textStyle.lineHeight = '40px';
```
danimoh: If Muli is available on the page, I think we should use it. I.e.:
`textStyle.fontFamily = 'Muli', fallbacks...`

PR #355, opened 2019-11-30
Reset password error code handling

nibhar:
This Pr makes use of the Error added in nimiq/keyguard#389 to implement the reset-password pseudo request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "dependencies": {
         "@nimiq/browser-warning": "^1.1.0",
         "@nimiq/iqons": "^1.5.0",
-        "@nimiq/keyguard-client": "^1.1.0",
+        "@nimiq/keyguard-client": "^1.2.0",
         "@nimiq/ledgerjs": "https://github.com/nimiq/ledger-api.git",
         "@nimiq/network-client": "^0.3.2",
         "@nimiq/rpc": "^0.4.0",
--- a/src/views/ErrorHandler.vue
+++ b/src/views/ErrorHandler.vue
@@ -61,6 +61,40 @@ export default class ErrorHandler extends Vue {
             return;
         }
 
+        if (this.keyguardResult.message === Errors.Messages.GOTO_RESET_PASSWORD) {
+            try {
+                const walletInfo = await this.getWalletForThisRequest();
+                if (walletInfo) {
+                    const request: KeyguardClient.ResetPasswordRequest = {
+                        appName: this.request.appName,
+                        requestedKeyPaths: [
+                            DEFAULT_KEY_PATH,
+                            ...[...walletInfo.accounts.values()].map((account) => account.path),
+                        ],
+                        isKeyLost: false,
+                        expectedKeyId: walletInfo.keyId,
+                        wordsOnly: true,
+                    };
+
+                    const client = this.$rpc.createKeyguardClient();
+                    client.resetPassword(request);
+
+                    return;
+                } else {
+                    // This is most definitely never going to happen.
+                    // It would require the initial CHANGE_PASSWORD request to go through (and find the wallet)
+                    // while the subsequent call to this function needs to then not find it anymore.
+                    // However in the unlikely scenario of it happening nonetheless we provide a proper error
+                    // and close the window.
+                    this.$rpc.reject(new Error('Wallet does not exist.'));
+                    return;
+                }
+            } catch (error) {
+                this.$rpc.reject(error);
+                return;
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
     "dependencies": {
         "@nimiq/browser-warning": "^0.4.0",
         "@nimiq/iqons": "^1.5.0",
-        "@nimiq/keyguard-client": "^1.0.0",
+        "@nimiq/keyguard-client": "link:../keyguard-next/client/",
```
sisou: Reminder to replace by new client package after Keyguard PR is merged.

src/views/ErrorHandler.vue
```diff
@@ -61,6 +61,32 @@ export default class ErrorHandler extends Vue {
...
+                        isKeyLost: false,
+                        expectedKeyId: walletInfo.keyId,
+                        wordsOnly: true,
+                    };
+
+                    const client = this.$rpc.createKeyguardClient();
+                    client.import(request);
+
+                    return;
+                }
```
sisou: I think there should be an `else` case (when wallet is not found (I know, unlikely but possible?), otherwise you can just use `walletInfo!` everywhere and don't need the `if (walletInfo`) check) where the situation is explained in the error message.

src/views/ErrorHandler.vue
```diff
@@ -61,6 +61,32 @@ export default class ErrorHandler extends Vue {
...
 
+        if (this.keyguardResult.message === Errors.Messages.RESET_PASSWORD) {
+            try {
+                const walletInfo = await this.getWalletForThisRequest();
+                if (walletInfo) {
+                    const request: KeyguardClient.ImportRequest = {
+                        appName: this.request.appName,
+                        requestedKeyPaths: [
+                            DEFAULT_KEY_PATH,
+                            ...[...walletInfo.accounts.values()].map( (account) => account.path),
```
sisou:
```suggestion
                            ...[...walletInfo.accounts.values()].map((account) => account.path),
```

src/views/LoginSuccess.vue
```diff
@@ -179,6 +179,13 @@ export default class LoginSuccess extends Vue {
             await new Promise((resolve) => { this.resolve = resolve; });
         }
 
+        if (this.request.kind === RequestType.CHANGE_PASSWORD ) {
```
```suggestion
        if (this.request.kind === RequestType.CHANGE_PASSWORD) {
```

src/views/ErrorHandler.vue
```diff
@@ -61,6 +61,32 @@ export default class ErrorHandler extends Vue {
             return;
         }
 
+        if (this.keyguardResult.message === Errors.Messages.RESET_PASSWORD) {
```
sisou:
```suggestion
        if (this.keyguardResult.message === Errors.Messages.GOTO_RESET_PASSWORD) {
```

src/views/ErrorHandler.vue
nibhar: Agreed, I added the else case and a comment explaining it.

sisou:
To fix the merge conflict, you need to recreate what you need from these removed lines:
https://github.com/nimiq/hub/commit/26630455e9b7bb14def463218cf9832e4e096298#diff-1a2f78b804432c7a2fd11f022ac3ae18

src/views/LoginSuccess.vue
```diff
@@ -176,6 +179,13 @@ export default class LoginSuccess extends Vue {
             await new Promise((resolve) => { this.resolve = resolve; });
         }
 
+        if (this.request.kind === RequestType.CHANGE_PASSWORD) {
```
danimoh: Might want to add a comment how we can end up from a `CHANGE_PASSWORD` request in `LoginSuccess`.

src/views/ErrorHandler.vue
```diff
@@ -61,6 +61,39 @@ export default class ErrorHandler extends Vue {
...
+                    client.import(request);
+
+                    return;
+                } else {
+                    // This is most definitely never going to happen.
+                    // It would require the initial CHANGE_PASSWORD request to go through (and find the wallet)
+                    // while the subsequent call to this function needs to then not find it anymore.
+                    // However in the unlikely scenario of it happening nonetheless we provide a proper error
+                    // and close the window.
+                    this.$rpc.reject(new Error('Wallet does not exist.'));
```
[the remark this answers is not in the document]
danimoh: Note that this comment has not been addressed yet.

PR #364, opened 2020-01-06
Cashlink Ledger Support

danimoh:
Adapted / rebased from https://github.com/nimiq/hub/pull/325


src/views/SignTransactionLedger.vue
```diff
@@ -1,8 +1,11 @@
 <template>
-    <div class="container">
+    <div v-if="request.kind !== 'cashlink' || !!cashlink" class="container">
```
sisou: I don't understand this condition? Always show when the request is NOT of type 'cashlink' or when there is a cashlink variable? How does this make sense? Especially since you still check for the 'cashlink' request type in the conditions below.

src/views/SignTransactionLedger.vue
```diff
@@ -17,18 +20,19 @@
                     :address="recipientDetails.address"
                     :label="recipientDetails.label"
                     :image="recipientDetails.image"
-                    @click.native="shownAccountDetails = recipientDetails"
+                    :displayAsCashlink="recipientDetails.isCashlink"
+                    @click.native="shownAccountDetails = !recipientDetails.isCashlink ? recipientDetails : null"
```
sisou: I overlooked the negating `!` at first and was confused, so I suggest to turn this condition around:
```suggestion
                    @click.native="shownAccountDetails = recipientDetails.isCashlink ? null : recipientDetails"
```

src/views/SignTransactionLedger.vue
```diff
@@ -85,18 +89,26 @@ import StatusScreen from '../components/StatusScreen.vue';
...
-import { ParsedCheckoutRequest, ParsedSignTransactionRequest, RequestType } from '../lib/RequestTypes';
+import {
+    ParsedCashlinkRequest,
+    ParsedCheckoutRequest,
+    ParsedSignTransactionRequest,
+    RequestType,
+} from '../lib/RequestTypes';
 import { WalletInfo } from '../lib/WalletInfo';
-import { ERROR_CANCELED, TX_VALIDITY_WINDOW, CASHLINK_FUNDING_DATA } from '../lib/Constants';
+import { CASHLINK_FUNDING_DATA, ERROR_CANCELED, TX_VALIDITY_WINDOW } from '../lib/Constants';
```
sisou: Alphabetical sorting? Is that necessary?

src/views/SignTransactionLedger.vue
```diff
@@ -153,8 +166,22 @@ export default class SignTransactionLedger extends Vue {
...
+            // coming from cashlink create
+            if (!this.cashlink) {
+                // happens if user reloads the page
+                this._back();
+                return;
+            }
+            senderUserFriendlyAddress = this.$store.state.activeUserFriendlyAddress;
+
+            this.recipientDetails = {
+                address: this.cashlink!.address.toUserFriendlyAddress(),
```
sisou: Since there is a `return` in the check for `this.cashlink` above, I think the `!` might not be necessary here anymore:
```suggestion
                address: this.cashlink.address.toUserFriendlyAddress(),
```

src/views/SignTransactionLedger.vue
```diff
@@ -153,8 +166,22 @@ export default class SignTransactionLedger extends Vue {
...
+                label: 'New Cashlink',
+                isCashlink: true,
+            };
         } else {
-            this.$rpc.reject(new Error('Must be invoked via sign-transaction or checkout requests.'));
+            this.$rpc.reject(new Error('Must be invoked via sign-transaction, checkout or cashlink requests.'));
```
sisou: Can you add to this error message _what_ must be invoked? Like maybe 'Ledger Transaction Signing must be invoked...'? That would be a more helpful error message by itself without having to look at the stack trace.

src/views/SignTransactionLedger.vue
```diff
@@ -175,17 +202,32 @@ export default class SignTransactionLedger extends Vue {
...
             const blockchainHeight = await network.getBlockchainHeight(); // usually instant as synced in checkout
             // The next block is the earliest for which tx are accepted by standard miners
             validityStartHeight = blockchainHeight + 1
                 - TX_VALIDITY_WINDOW
                 + checkoutRequest.validityDuration;
+        } else if (this.request.kind === RequestType.CASHLINK) {
+            if (!this.cashlink) {
+                // already handled in created
+                return;
+            }
```
sisou: If it's already handled, why is this necessary here? Only for TS to not complain?

src/views/SignTransactionLedger.vue
```diff
@@ -201,19 +243,19 @@ export default class SignTransactionLedger extends Vue {
...
+                extraData: data,
+                flags,
             }, signer.path, senderAccount.keyId);
         } catch (e) {
             // If cancelled, handle the exception. Otherwise just keep the error message displayed in ledger ui.
             if (e.message.toLowerCase().indexOf('cancelled') !== -1) {
-                if (this.request.kind === RequestType.CHECKOUT) {
-                    this._back(); // user might want to choose another account or address
+                if (this.request.kind === RequestType.CHECKOUT || this.request.kind === RequestType.CASHLINK) {
+                    this._back(); // user might want to choose another account or address or change cashlink
```
sisou: Does going back to the cashlink form have the values prefilled again? Or do the value and message get lost when going back?

src/lib/Cashlink.ts
```diff
@@ -269,16 +281,18 @@ class Cashlink {
     }
 
     public getFundingDetails(): {
-        layout: SignTransactionRequestLayout,
-        recipient: Uint8Array,
+        layout: 'cashlink',
+        recipient: Nimiq.Address,
```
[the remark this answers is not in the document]
danimoh: I changed it because in the new Keyguard client the SignTransactionRequest is typed more strictly with incompatible types for different layout. Therefore, the layout also needs to be specified more strictly here. While the Keyguard client is not updated in this PR, eventually this change will be necessary.
I'll remove the import for `SignTransactionLayout`.

The recipient type I changed to be easier to use in `SignTransactionLedger`.

src/views/CashlinkManage.vue
```diff
@@ -93,20 +93,33 @@ export default class CashlinkManage extends Vue {
...
-                this.$rpc.reject(new Error(`${this.request.cashlinkAddress} is not a valid Cashlink.`));
+        let storedCashlink;
+        if (this.request.cashlinkAddress) {
+            storedCashlink = await CashlinkStore.Instance.get(this.request.cashlinkAddress.toUserFriendlyAddress());
+            if (!storedCashlink) {
+                this.$rpc.reject(new Error(`Could not find Cashlink for address ${this.request.cashlinkAddress}.`));
                 return;
             }
+        } else if (this.cashlink) {
+            storedCashlink = await CashlinkStore.Instance.get(this.cashlink.address.toUserFriendlyAddress());
```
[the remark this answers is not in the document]
danimoh: `storedCashlink` always contains a cashlink from the cashlink store, which can be null, even if `this.cashlink` is defined. They are thus not the same.

src/views/CashlinkManage.vue
```diff
@@ -93,20 +93,33 @@ export default class CashlinkManage extends Vue {
...
+        }
+
+        if (storedCashlink) {
+            // cashlink already sent and stored
+            this.retrievedCashlink = storedCashlink;
+            this.isTxSent = true;
+            this.isManagementRequest = !this.cashlink; // freshly created cashlink or management request?
+        } else {
+            this.retrievedCashlink = this.cashlink!;
         }
```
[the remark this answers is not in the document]
danimoh: That's not correct. See comment above.

src/views/SignTransactionLedger.vue
danimoh: If the `request.kind` is `'cashlink'` the template should only be rendered, if `cashlink` is defined to avoid accessing the properties on the cashlink if it's undefined.
I can reformulate this to `request.kind === 'cashlink' ? !!cashlink : true".

src/views/SignTransactionLedger.vue
danimoh: My IDE does that automatically, so why not.

src/views/SignTransactionLedger.vue
danimoh: `mounted` can execute before the `reject` closes the window. Therefore I need to stop the execution here.

src/views/SignTransactionLedger.vue
danimoh: They're lost.

src/views/SignTransactionLedger.vue
sisou: Don't know if that's better :D I did not find the condition formulation confusing, just that it's there in the first place :stuck_out_tongue:

src/views/SignTransactionLedger.vue
sisou: Could you put that in the comment at the next opportunity? :slightly_smiling_face:

src/views/SignTransactionLedger.vue
sisou: Understood. So when does this `cancelled` error happen? When the user cancels the action on the Ledger itself? Or by clicking a cancel button in the UI? What happens if the Ledger times out, does it then also do this and go back to the cashlink route?

src/lib/Cashlink.ts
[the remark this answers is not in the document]
sisou: >While the Keyguard client is not updated in this PR
There you go. So the change has nothing to do with the Ledger support. Just wanted to make sure :wink: 

Does the new Keyguard client export the layouts as values to use here?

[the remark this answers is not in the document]
danimoh: The layout is not an enum in the types but just a union over the constant strings (I don't know why).
Therefore this is as good as we can do it, I guess.

src/views/SignTransactionLedger.vue
danimoh: The changes for checkout/ledger move all the logic to the `mounted` hook which makes this clearer.

src/views/SignTransactionLedger.vue
danimoh: When cancelled on the ledger or when cancelled via code which is triggered, when the view gets destroyed.
I added a change that disables handling the error when the view already is destroyed.
This did not cause any problems so far, but it's cleaner to handle this.

sisou: Ok. My problem is then that when I cancel it on the Ledger, instead of being able to directly retry, all my Cashlink form input is lost and I have to enter the value and message again. Correct?

I would rather that when I cancel on the Ledger, I get a retry screen and only when I click the arrow back on that screen, I get back to the Cashlink form.

danimoh: But why would a user cancel but then still want to sign the same transaction?

The proper solution would be that the cashlink create form should remember the data when coming back from ledger or keyguard.

sisou: Yes, you are propably right, if there is no additional user-action between clicking the "create cashlink" button and the start of the Ledger process.

Making the form remember the values would indeed be preferrable, I agree. Should even be possible, because the cashlink is stored in the static store, right?

src/lib/Cashlink.ts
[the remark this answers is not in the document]
nibhar: The export of these constants we should change in the KeyguardClient.

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
nibhar: After the keyguard returns to this code, the cashlink will first enter the `CashlinkStore` a couple of lines down in Line 144. Thus @danimoh is correct in saying that `storedCashlink` will be `null` here in that case.

However, if the browser crashes during sending (which is why the storing part was moved up) and the page is reloaded it will no longer be `null`, but a transaction was not send.

src/views/SignTransactionLedger.vue
```diff
@@ -153,8 +167,23 @@ export default class SignTransactionLedger extends Vue {
                 label: this.rpcState.origin.split('://')[1],
                 image: checkoutRequest.shopLogoUrl,
             };
+        } else if (this.request.kind === RequestType.CASHLINK) {
```
nibhar: I guess these `if`s can now be a `switch`.

danimoh:
I pushed an additional change that I wanted to make to restore cashlinks from the static store on reload.
The previous handling on reload was to navigate back to the cashlink creation screen, even when on the CashlinkManage screen after the cashlink was already successfully created and sent, which was not the best user experience.
I implemented this as a general change, that makes the staticStore contents available to page reloads. This also fixes issues on reload in other flows, for example the `originalRoute` not being available anymore after reload.
Please look for undesired side effects.
I think the changes should be safe, as they only come to effect on reloads and for these generally the behavior probably rather improved.

I also made the restoration of the staticStore on back navigation from the Keyguard consistent, regardless of whether `handleHistoryBack` / `rejectOnBack` is set or not.

I then also used this to restore the inputs in CashlinkCreate on navigation back from the Keyguard or Ledger Signing.

src/views/SignTransactionLedger.vue
danimoh: I wouldn't do it now to avoid unnecessary merge conflicts with the checkout/ledger changes as using a switch instead of if/else changes the indentation of the code.
For this reason I also don't necessarily prefer a `switch` here.

src/views/SignTransactionLedger.vue
danimoh: The form does remember the values now.

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
danimoh: The main target to move the storing up was to avoid funds getting lost by a transaction that was sent to the network without storing the Cashlink. This is still met by the current implementation.

Additionally, harm to the user experience / user confusion is limited as the Cashlink will not pop up in the safe as there is no matching transaction. When he then sees that his balance did not change and there is no Cashlink charging transaction, he can reason, that no Cashlink was created and attempt to do so again.

In the previous implementation, the transaction signature was available after reloads because the keyguard result is written to session storage in the rpc client. This might or might not be available after a browser crash depending on whether the user configured his browser to resume sessions and whether the browser already wrote that information to disk. Thus, also the previous implementation was not completely fail safe.

However if the transaction failed to be sent without the browser crashing, the previous implementation was able to resend the transaction on reload.
I could archive a similar effect by storing the signed transaction in the history state.

Alternatively, we could think about storing transactions that have not been sent to the network yet in localstorage and then trying to send them out to the network whenever the hub is opened with a request that stays in the hub (i.e. does not simply redirect to the keyguard). That would be a more sophisticated solution as it would also work after the window is closed, also for transmission of checkout transactions.
Transactions that reach the end of their validity duration would get pruned.
On the other hand, this might feel confusing to the user if a transaction that failed in his opinion is later sent to the network.
Thus, maybe keeping the tx only until the current tab is closed, is actually preferrable?

src/views/SignTransactionLedger.vue
nibhar: I would prefer a `switch` here. If it leads to merge conflicts, it can be done in the next PR touching this though.

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
nibhar: I would agree that transaction should be send in the flow they are created in, not at some unspecified and unreliable point in time in the future.

I wanted to point out that there is a possibility, that this condition fails, as there is a disconnect in when a Cashlink enters the store and when it becomes functional. I am aware that no funds can be lost.

I agree that the harm done to the user is small and more an inconvenience than harm. 
However alerting the user to this inconvenience and not waiting for him to find out on its own might be preferable here. Could potentially be done in a separate PR, as it is not Ledger specific since, as you pointed out, the previous implementation was not fail-safe as well.

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
nibhar: So this would be expected to result in `storedCashlink === null` in case of a Keyguard signed Cashlink and `storedCashlink !== null` in the ledger case?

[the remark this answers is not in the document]
danimoh: Yes, exactly.

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
danimoh: I agree. Should be a separate PR to get this one merged faster.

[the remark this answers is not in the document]
sisou: So this code fails in case the transaction sending fails/hangs/crashes the browser and the user reloads? Then it will look like it sent? Does it also look like it sent when I relead the page just as the consensus gets established before the tx sends?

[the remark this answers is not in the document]
nibhar: I want to say yes.

src/lib/Cashlink.ts
```diff
@@ -110,12 +123,14 @@ class Cashlink {
...
     private _messageBytes: Uint8Array = new Uint8Array(0);
     private _value: number | null = null;
+    private _fee: number | null = null;
     private _knownTransactions: DetailedPlainTransaction[] = [];
 
     constructor(
         public keyPair: Nimiq.KeyPair,
         public address: Nimiq.Address,
         value?: number,
+        fee?: number,
```
sisou: Optional parameters usually come at the end of the argument list, to not have to write `undefined` in the middle of the function call / object creation...

src/lib/Cashlink.ts
```diff
@@ -128,6 +143,7 @@ class Cashlink {
         this._getNetwork = () => networkPromise;
 
         if (value) this.value = value;
+        if (fee) this._fee = fee; // set via _fee to avoid throwing an exception in fee setter if already claimed
```
sisou: Why add a check for immutability and then go around it here? I think this is bad and the immutability check above should just be a warning then.

src/lib/CashlinkStore.ts
```diff
@@ -30,7 +30,11 @@ export class CashlinkStore extends Store<Cashlink, CashlinkEntry> {
     }
 
     protected toEntry(cashlink: Cashlink): CashlinkEntry {
-        return cashlink.toObject();
+        const entry = cashlink.toObject();
+        // delete fee and currently unused contactName which are unnecessary to store
+        delete entry.fee;
+        delete entry.contactName;
+        return entry;
```
sisou: Is a cashlink's `.toObject()` used anywhere currently where the fee and contactName is used? If not, I suggest to remove the fee and contactName from the return value of `toObject` be done with it. Would also avoid the abomination of a function name below (`expectEqualIgnoringFeeAndContactName`).

src/lib/Cashlink.ts
danimoh: All following parameters are optional.

src/lib/Cashlink.ts
danimoh: This is to be able to (re)create cashlink instances (for example from the store) that are already claimed and have a fee assigned.
The sanity check is to block changing the fee when it wouldn't have any effect anymore.
The construction however is an atomic operation at which all this is an allowed configuration.

sisou:
I have several problems here in how you are handling this PR:

1. Why do you need to completely change code that we all agreed on already in a way that adds new bugs (CashlinkManage.vue, the long conversation above). Couldn't you just add an additional condition for the Ledger case without changing the whole thing of how cashlink are stored, loaded, interpreted? I would prefer that.
2. I disagree with your last commits of adding a whole change of how the static store is stored in history etc, only because of the small comment that the cashlink for does not refill values. This new feauture has nothing to do with the Ledger changes and adding such a massive change to this PR is a horrible idea, especially after someone already reviewed the PR and we talked about stuff. Please remove all commits related to staticStore changes.

We need to keep PRs small and **focused** or otherwise we will be here discussing this stuff until next christmas!

src/lib/CashlinkStore.ts
danimoh: `toObject` should return an object representation of the cashlink from which it can be recreated.
To omit data there just to avoid it being included in an independent class would be a hack.
The data is used to recreate cashlinks on reload in the ledger flow and to restore the inputs in `CashlinkCreate`.

src/lib/Cashlink.ts
sisou: I think that actually setting a fee after the cashlink was claimed should not be an error. It can be a warning, but why should it throw?

danimoh:
1. The code has to be changed to support Ledgers. I didn't hear your smart suggestion yet how you would have done this.
2. These changes are directly related to the Ledger PR as they enable page reloads during Ledger cashlink signing. This change was not because of the form refilling. Form refilling just happened to be able to use the same logic. (See my previous comment).

Right now I'm wasting time answering an unnecessary comment.

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
danimoh: The consensus is already established in `SignTransactionLedger`.
And we already said we will handle this in a separate PR.
I will resolve this conversation now.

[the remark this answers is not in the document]
sisou: I'm not talking about Ledger-signed cashlinks, I'm talking about Keyguard-signed cashlinks. It sounds to me like you **introduced** a bug, where when I come back from the Keyguard after entering my password and I get the loading screen in the Hub here that *then establishes consensus here* and tries to send the funding transaction, and for any reason stops working and I reload the page, then it immediately shows me that the cashlink was funded successfully, although no transaction was ever sent.

Is my understanding correct? If so, I disapprove this PR because that is a serious new bug for me.

sisou:
You **added** the whole staticStore persistence **after** you published the PR, so it could not have been necessary for the Ledger-related changes that I reviewed first, otherwise it would have been already included. I am saying the commits starting with "RpcApi: export state to history" should be removed from this PR because they are a separate feature.

danimoh:
As already pointed out the reload enabling changes belong to this PR.
I pushed the initial version early for you guys to be able to start reviewing earlier.
I could postpone the two commits starting at `CashlinkCreate: restore values on back navigation` for a separate PR but I don't see much value in that, especially as this now already went through review.

src/views/ExportSuccess.vue
```diff
@@ -83,8 +83,7 @@ export default class ExportSuccess extends Vue {
 
             // Recreate original URL with original query parameters
             const query = { rpcId: staticStore.rpcState!.id.toString() };
-            setTimeout(
-                () => this.$router.push({ name: RequestType.MIGRATE, query }), StatusScreen.SUCCESS_REDIRECT_DELAY);
+            setTimeout(() => this.$rpc.routerPush(RequestType.MIGRATE, query), StatusScreen.SUCCESS_REDIRECT_DELAY);
```
nibhar: Looks to me as if we might need to take action to prevent us using `this.$router.push()` and `this.$router.replace()` as they expose a behaviour which is no longer desirable?

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
nibhar: Actually not sure how to best proceed here.
I kind of agree with @sisou that the behaviour especially for a reload should be working. However there is also a huge discussion about what to include and not include going on.

I personally would probably opt to restore the reload behaviour as it was working before, while moving the rest of the problem to a different PR. I assume that i.e. a crashed browser is sort of a good indicator, that something went wrong as well and handling that case could be postponed until later. Especially since no funds would be lost.

src/lib/Cashlink.ts
```diff
@@ -27,22 +27,33 @@ class Cashlink {
...
+            throw new Error('Cannot set value, Cashlink is immutable');
         }
-        if (!Nimiq.NumberUtils.isUint64(value) || value === 0) throw new Error('Malformed value');
+        if (!Nimiq.NumberUtils.isUint64(value) || value === 0) throw new Error('Malformed Cashlink value');
         this._value = value;
     }
 
+    set fee(fee: number) {
+        if (this.state === CashlinkState.CLAIMED) {
+            throw new Error('Cannot set fee, Cashlink is already claimed');
```
sisou: I think it should not be an error/exception when a fee is set after it is useful, as the fee - different to the value and the message - has no effect on the cashlink url (which is the reason why value and message are immutable).

src/lib/CashlinkStore.ts
sisou: In my opinion the cashlink's `.toObject()` should omit the fee (and maybe contactName, as long as it's unused) in the first place. The fee is so unimportant to the cashlink itself that it does not need to be exported.

src/lib/RpcApi.ts
```diff
@@ -98,17 +103,29 @@ export default class RpcApi {
             undefined, // preserveRequests: keep default behavior, which is true for redirects but false for postMessage
             handleHistoryBack,
         );
+        if (!handleHistoryBack) {
+            // The Keyguard client rejects on history back only if handleHistoryBack is activated. If the Keyguard does
+            // not reject it also does not provide us the localState to recover. For this case, we encode it manually in
+            // the history, to retrieve it from there.
+            this._exportStateToHistory();
+        }
```
sisou: This assumes that the Keyguard client is always created just-in-time when it's invoked. This is currently the case, but not enforced. So this might export an incomplete state to the history, when changes to the state are done after the Keyguard client is created via this function.

src/lib/RpcApi.ts
```diff
@@ -98,17 +103,29 @@ export default class RpcApi {
...
     }
 
-    public routerPush(routeName: string) {
-        const query = this._parseUrlParams(window.location.search);
-        this._router.push({name: routeName, query});
+    public routerPush(routeName: string, query?: { [key: string]: string }) {
+        query = query || this._parseUrlParams(window.location.search);
+        this._router.push({name: routeName, query}, () => {
+            // export state to the newly pushed history entry to be available on reload
+            this._exportStateToHistory();
```
sisou: So spamming the browser's history storage with JSON objects on every navigation is required to be able to reload at all times?

src/lib/RpcApi.ts
sisou: Furthermore, this now stores the very same data as the KeyguardClient stores anyway (in localstorage I believe) in a second location. Is that necessary? Can we get rid of the duplication? Does anyone remember why the RPC client is not storing its `localState` in the history object but in localstorage instead?

src/lib/RpcApi.ts
```diff
@@ -167,6 +184,10 @@ export default class RpcApi {
         };
     }
 
+    private _exportStateToHistory() {
+        history.replaceState({ ...history.state, exportedState: this._exportState() }, '');
+    }
```
sisou: Would it also be possible to not store in the history object, but in localstorage and replace it there on every navigation? Would that be able to fulfill the same purpose?

src/lib/RpcApi.ts
sisou: Also, the callback is the `onComplete` callback here and gets called after the navigation is complete. I believe it makes more sense to export the state _before and after_ the navigation: the latest state still in the previous history entry - and also first thing in the new screen, later being replaced by the latest state when navigating away. That way the latest state, and all the changes made to the state in that screen, are preserved in its history entry, not in the next. 

Right now this saves a state to the history, which then changes in the screen (because that's what the UI is for, to change the state), but that state then only gets saved in the _next_ navigation entry.

Am I thinking right?

src/lib/Cashlink.ts
danimoh: Throwing is more consistent to what the `value` and `message` setters do.
Equally one could discuss whether the `value` and `message` setters should rather warn than throw as changing the value can make sense for cashlinks that should get recharged or be able to be claimed multiple times with smaller amounts. Also changing the cashlink message after creation can still make sense, for example in the gift card app, the cashlink message in the QR code could get updated when the message in the gift card gets changed.
Given these aspects, throwing for `value` and `message` but not for `fee` feels arbitrary to me.
However, in the end, I don't care too much whether it throws or warns.
To get this merged, I changed it.

src/lib/Cashlink.ts
danimoh: See discussion in https://github.com/nimiq/hub/pull/364#discussion_r366049491

src/lib/CashlinkStore.ts
danimoh: The fee was added to `toObject` when it became necessary to have it there.
It is used for recreating a Cashlink from the history store to the static store on reload in `CashlinkCreate` and `SignTransactionLedger`.
I don't see why the cashlink lib should not export these values which are a part of describing a cashlink instance and are needed to recreate an equivalent instance.
To make things slightly nicer in the `CashlinkStore` is not really a valid reason.
A compromise would be to add a flag to `toObject` that omits the `fee` and `contactName`.

src/views/ExportSuccess.vue
danimoh: What would you suggest, removing the `$router` from our vue instances or overwriting `router.push` to print a warning?

src/lib/RpcApi.ts
danimoh: The history state is exported exactly at the time the Keyguard Client is created with the rpc state.
To achieve this consistency was the idea of this change.

The data is duplicated yes. But it is only stored in the history or sessionStorage, i.e. no persistent data is duplicated. That's ok in my opinion.
Also the history state and Keyguard rpc state become relevant in different situations:
- The keyguard state is relevant when returning from the Keyguard after a successful or failed request.
- The history state is relevant for enabling reloading of pages within the Hub without loosing the exported state. This is relevant for several requests, for example Ledger requests that don't go through the Keyguard but also other requests that make use of the static store like requests that store an originalRoute.
- The history state is also used to restore the state when coming back from the Keyguard by a simple back navigation without that the Keyguard resolves or rejects the Keyguard request in which case the Keyguard does not provide the state. The goal is thus to be able to restore the state regardless of whether `handleHistoryBack` is set or not. This third bullet point is what the lines that this comment has been attached to achieve.

The Keyguard / rpc lib stores the state in the sessionStorage as it needs to be able to restore the data after a Keyguard requests finished, which will be another history entry than the one that invoked the Keyguard.

src/lib/RpcApi.ts
danimoh: I don't think that the use of either `localStorage` (or `sessionStorage`) would be recommendable.
`localStorage` would be persistent which we don't need and would thus be more wasteful. `localStorage` is also shared across multiple tabs which potentially leads to problems.
When using `sessionStorage` we only have one store instance we'd be overwriting all the time, thus on back navigation we can't restore the state we had at that specific page / history entry. This becomes for example problematic when executing multiple redirect requests in a row in the same tab. If the user navigates back to the previous request, we wouldn't be able to resume the state for that request (if it doesn't come from other sources like the request).

We might be able to use the session storage if accessing the state in the session storage by request id. But we wouldn't really win anything by doing so and the use of the history state is more general and cleaner.

src/lib/RpcApi.ts
danimoh: See https://github.com/nimiq/hub/pull/364#discussion_r366095837 for an answer related to your first question.
Firefox imposes a limit of 640k characters per history state and we're way below that limit, so I guess we're not super wasteful here.

The question about when to export the state to the history is a design decision.
Storing stuff in the static store is mostly done as input for the _next_ page. Therefore the data is only persisted in the state for the next page.
We can discuss whether there are use cases for updating the store also for the current page.

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
sisou: I think the case where a cashlink funding transaction fails to send because consensus-establishing fails (after coming back from Keyguard) is way more common than creating a Cashlink with Ledger. So I would optimize for doing the expected thing when reloading the consensus-loading-screen.

src/views/ExportSuccess.vue
sisou: I think removing `this.$router` is not recommended, because it's set internally by vue-router when the plugin is activated, and likely also used internally by the plugin.

I think it makes sense to add a navigation guard in dev mode (https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards), which checks that the navigation was done with the correct method and warn (or even block) in dev mode.

src/lib/RpcApi.ts
sisou: The idea of exporting to history state grew on me in the last days. I'm accepting it.

src/lib/Cashlink.ts
```diff
@@ -103,21 +116,24 @@ class Cashlink {
      * Cashlink balance in luna
      */
     public balance: number | null = null;
+    public state: CashlinkState;
```
sisou: I know we said we are not going to comment on code style much anymore, but this is a change that really grinds my gears again. There is no reason that the `state` needs to be changed from being indirectly assigned in the constructor arguments to this.

src/lib/Cashlink.ts
```diff
@@ -103,21 +116,24 @@ class Cashlink {
...
     private _knownTransactions: DetailedPlainTransaction[] = [];
 
     constructor(
         public keyPair: Nimiq.KeyPair,
         public address: Nimiq.Address,
         value?: number,
+        fee?: number,
         message?: string,
-        public state: CashlinkState = CashlinkState.UNCHARGED,
+        state: CashlinkState = CashlinkState.UNCHARGED,
```
sisou: Together with comment above: why?

src/lib/Cashlink.ts
```diff
@@ -128,7 +144,9 @@ class Cashlink {
         this._getNetwork = () => networkPromise;
 
         if (value) this.value = value;
+        if (fee) this.fee = fee;
         if (message) this.message = message;
+        this.state = state;
```
sisou: Together with comment above: why?

src/lib/Cashlink.ts
danimoh: Typically every change has a reason ;)
This change enables setting the fee via the `fee` setter in the constructor instead of using `_fee` without that the warning gets triggered.

sisou: Because `state` needs to be assigned _after_ the `fee` to not trigger the warning. I get it, but it's not nice :(

src/views/CashlinkManage.vue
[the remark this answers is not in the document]
danimoh: Resending of unrelayed transactions is addressed by https://github.com/nimiq/hub/pull/368

src/lib/RpcApi.ts
danimoh: I discussed this here with Sebastian and we decided that for the moment the current implementation makes more sense:
- The static store is mostly used to set input for the next page.
- Restoring the static store while other parts of the page are not restored (inputs, etc.) might be error prone or lead to confusion by the user.
- As the static store is not reactive, restoring a state might be undesirable as later changes will not lead to reactive updates.

src/views/ExportSuccess.vue
nibhar: Could we remove the `router_replace/push` methods from RpcApi in favour of the navigation guards? 
If not this still seems to me like the preferred solution as well.

danimoh: Good idea :+1:
I opened an issue for this: https://github.com/nimiq/hub/issues/371

PR #368, opened 2020-01-14
Unrelayed tx caching

danimoh:
This PR enables resending transactions after reload of the page.
This has been a feature of `CashlinkManage` for Keyguard Cashlink transactions before but is now also supported for Ledger transactions (checkout and cashlink).
Additionally, `CashlinkManage` can now detect for Keyguard transactions whether they have already been sent or not, and does now not resend a transaction on reload anymore but forwards directly to the management screen.

### DIFF
--- a/src/components/Network.vue
+++ b/src/components/Network.vue
@@ -130,7 +130,7 @@ class Network extends Vue {
         }, '');
 
         const signedTx = await this.makeSignTransactionResult(tx);
-        const client = await this._getNetworkClient();
+        const client = await this.getNetworkClient();
 
         const txObjToSend = Object.assign({}, signedTx.raw, {
             senderPubKey: signedTx.raw.signerPublicKey,
@@ -184,24 +184,24 @@ class Network extends Vue {
     }
 
     public async getBlockchainHeight(): Promise<number> {
-        const client = await this._getNetworkClient();
+        const client = await this.getNetworkClient();
         if (Network._hasOrSyncsOnTopOfConsensus) return client.headInfo.height;
         return new Promise((resolve) => this.$once(Network.Events.CONSENSUS_ESTABLISHED,
             () => resolve(client.headInfo.height)));
     }
 
     public async getBalances(addresses: string[]): Promise<Map<string, number>> {
-        const client = await this._getNetworkClient();
+        const client = await this.getNetworkClient();
         return client.getBalance(addresses);
     }
 
     public async requestTransactionReceipts(address: string): Promise<Nimiq.TransactionReceipt[]> {
-        const client = await this._getNetworkClient();
+        const client = await this.getNetworkClient();
         return client.requestTransactionReceipts(address);
     }
 
     public async getGenesisVestingContracts(): Promise<VestingContractInfo[]> {
-        const client = await this._getNetworkClient();
+        const client = await this.getNetworkClient();
         const contracts = await client.getGenesisVestingContracts();
 
         return contracts.map((contract) => new VestingContractInfo(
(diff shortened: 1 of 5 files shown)


src/components/Network.vue
```diff
@@ -125,32 +135,65 @@ class Network extends Vue {
...
+        value?: number,
+        fee?: number,
+        validityStartHeight?: number,
+        flags?: number,
+        data?: Uint8Array,
+    }): Nimiq.Transaction[] {
+        if (!getHistoryStorage(Network.HISTORY_KEY_UNRELAYED_TRANSACTIONS)) return [];
+        const seializedTransactions: Uint8Array[]
+            = Object.values(getHistoryStorage(Network.HISTORY_KEY_UNRELAYED_TRANSACTIONS));
+        const transactions = seializedTransactions.map((serializedTx: Uint8Array) =>
```
sisou: Typo:

```suggestion
        const transactions = serializedTransactions.map((serializedTx: Uint8Array) =>
```

src/lib/RpcApi.ts
```diff
@@ -28,8 +28,13 @@ import CookieJar from '@/lib/CookieJar';
 import { Raven } from 'vue-raven'; // Sentry.io SDK
 import { ERROR_CANCELED } from './Constants';
 import Config from 'config';
+import { setHistoryStorage, getHistoryStorage } from '@/lib/Helpers';
 
 export default class RpcApi {
+    private static get HISTORY_KEY_RPC_STATE() {
+        return 'rpc-state';
```
sisou: Careful with naming here, since the RPC lib internally also has a class called `State`, which is sometimes imported as `RpcState`, but here in the `RpcApi` class the State means something more and stores all of the Hub state. So I suggest calling it `hub-state`.

src/views/CashlinkManage.vue
```diff
@@ -108,55 +112,56 @@ export default class CashlinkManage extends Vue {
...
+            transactionToSend = network.getUnrelayedTransactions({
+                recipient: storedCashlink.address,
+                value: storedCashlink.value,
+            })[0];
+            this.isTxSent = !transactionToSend;
             this.retrievedCashlink = storedCashlink;
-            this.isTxSent = true;
-            this.isManagementRequest = !this.cashlink; // freshly created cashlink or management request?
         } else {
+            // Cashlink can not have been sent yet because whenever it gets sent, it also has been added to the store.
```
sisou: Just a suggestion:

```suggestion
            // Cashlink can not have been sent yet because whenever it gets sent, it was also added to the store.
```

src/views/CashlinkManage.vue
```diff
@@ -108,55 +112,56 @@ export default class CashlinkManage extends Vue {
...
+                recipient:  this.keyguardRequest.recipient,
+                recipientType: this.keyguardRequest.recipientType,
+                value: this.keyguardRequest.value,
+                fee: this.keyguardRequest.fee,
                 validityStartHeight: this.keyguardRequest.validityStartHeight,
+                data: this.keyguardRequest.data,
+                signerPubKey: this.keyguardResult.publicKey,
                 signature: this.keyguardResult.signature,
-                extraData: this.keyguardRequest.data,
-            });
```
sisou: Seems like this whole block can be shortened with object spreading:

```js
transactionToSend = transactionToSend || await network.createTx({
    ...this.keyguardRequest,
    signerPubKey: this.keyguardResult.publicKey,
});

danimoh: :+1:

danimoh:
@nibhar 
I think getting rid of the `Network` component is definitely still the idea.
However, at this moment, I did not want to also start messing with the `NetworkClient`.

PR #369, opened 2020-01-14
Small network fixes

danimoh:
Fixes two issues:
- Transaction relayed event listeners previously were removed even when it was another tx that was relayed. This way, `sendToNetwork` was not able to resolve anymore for one of the transactions, when two are sent at the same time.
- Previously, `getBlockchainHeight` reported one if called, before the consensus was established.

### DIFF
--- a/src/components/Network.vue
+++ b/src/components/Network.vue
@@ -134,14 +134,16 @@ class Network extends Vue {
         });
         client.relayTransaction(txObjToSend);
 
-        return new Promise<SignedTransaction>((resolve, reject) => {
-            this.$once('transaction-relayed', (txInfo: any) => {
+        return new Promise<SignedTransaction>((resolve) => {
+            const listener = (txInfo: any) => {
                 if (txInfo.hash !== base64Hash) return;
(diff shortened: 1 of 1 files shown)


src/components/Network.vue
```diff
@@ -179,7 +181,7 @@ class Network extends Vue {
         const client = await this.getNetworkClient();
         if (Network._hasOrSyncsOnTopOfConsensus) return client.headInfo.height;
         return new Promise((resolve) => this.$once(Network.Events.CONSENSUS_ESTABLISHED,
-            () => resolve(client.headInfo.height)));
+            () => setTimeout(() => resolve(client.headInfo.height), 10)));
```
sisou: 10 milliseconds is a bit arbitrary, don't you think? This is basically to wait a bit until the HEAD_CHANGED event fired after CONSENSUS_ESTABLISHED, yes? Do you think it makes sense to test for a hight > 1 every couple milliseconds and then resolve instead?

danimoh: Actually the `HEAD_CHANGED` event itself can be used which is being done now.

sisou: Ha, thought so :D perfect!

PR #370, opened 2020-01-14
Cashlink advanced third party api, including theme

danimoh:
Extend the cashlink create request by additional options:
- prefilled value
- prefilled message, with option automatically truncate if too long
- option to return the cashlink to the calling app with additional option to skip the sharing screen
- Cashlink theme

This also adds the Christmas theme.

This is currently marked as a draft and will be rebased later, but is already ready for review.


src/lib/RequestParser.ts
```diff
@@ -148,28 +152,75 @@ export class RequestParser {
...
+                    // CashlinkTheme enum and to filter out values that are actually the enum keys that have been added
+                    // by typescript for reverse mapping.
+                    throw new Error('Invalid Cashlink theme');
                 }
+
+                const returnCashlink = !!createCashlinkRequest.returnCashlink;
+                if (returnCashlink && !isPriviledgedOrigin(state.origin)) {
+                    throw new Error(`Origin ${state.origin} is not authorized to request returnCashlink.`);
+                }
+                const skipSharing = !!createCashlinkRequest.returnCashlink && !!createCashlinkRequest.skipSharing;
```
nibhar: Only a suggestion.

```suggestion
                const skipSharing = returnCashlink && !!createCashlinkRequest.skipSharing;
```

danimoh: It's actually not possible to do that simplification here, because `createCashlinkRequest.skipSharing` only exists if `createCashlinkRequest.returnCashlink` is defined.
Unfortunately ts does not recognize that `returnCashlink` is the same as `createCashlinkRequest.returnCashlink`.

src/lib/Helpers.ts
```diff
@@ -5,6 +5,7 @@ import {
     NETWORK_MAIN,
     ERROR_INVALID_NETWORK,
 } from '../lib/Constants';
+import { Utf8Tools } from '@nimiq/utils';
```
sisou: Unused.

src/lib/PublicRequestTypes.ts
```diff
@@ -139,20 +130,58 @@ export interface RenameRequest extends SimpleRequest {
...
+    value: number;
     status: CashlinkState;
-}
-
-export interface CashlinkRequest extends BasicRequest {
-    senderAddress?: string;
-    senderBalance?: number;
-    cashlinkAddress?: string;
+    theme: CashlinkTheme;
+    cashlink?: string;
```
sisou: I'd call this `url` or `link`, but not `cashlink`, because the container is already called `Cashlink`. This now means that users of this type have to write `cashlink.cashlink` in their code.

src/lib/RequestParser.ts
```diff
@@ -148,28 +152,75 @@ export class RequestParser {
...
+
+                let message = 'message' in createCashlinkRequest ? createCashlinkRequest.message : undefined;
+                if (message !== undefined) {
+                    if (typeof message !== 'string') {
+                        throw new Error('Cashlink message must be a string');
+                    }
+                    const { result: truncated, didTruncate } = Utf8Tools.truncateToUtf8ByteLength(message, 255);
+                    if (!('autoTruncateMessage' in createCashlinkRequest && createCashlinkRequest.autoTruncateMessage)
+                        && didTruncate) {
+                        throw new Error('Cashlink must be shorter than 256 bytes or autoTruncateMessage must be '
```
sisou: The word `message` is missing in the error message: "Cashlink **message** must be shorter than 256 bytes..."

src/lib/RequestParser.ts
```diff
@@ -148,28 +152,75 @@ export class RequestParser {
...
+                    skipSharing,
+                } as ParsedCreateCashlinkRequest;
+            case RequestType.MANAGE_CASHLINK:
+                const manageCashlinkRequest = request as ManageCashlinkRequest;
+                return {
+                    kind: RequestType.MANAGE_CASHLINK,
+                    appName: manageCashlinkRequest.appName,
+                    cashlinkAddress: manageCashlinkRequest.cashlinkAddress
+                        ? Nimiq.Address.fromUserFriendlyAddress(manageCashlinkRequest.cashlinkAddress)
+                        : undefined,
```
sisou: In the request type, `cashlinkAddress` is not optional. So I think this should throw when no address to manage is given here. Or what is the use case?

src/lib/RequestParser.ts
```diff
@@ -191,18 +242,29 @@ export class RequestParser {
...
                 return {
-                    appName: cashlinkRequest.appName,
-                    cashlinkAddress: cashlinkRequest.cashlinkAddress
-                            ? cashlinkRequest.cashlinkAddress.toUserFriendlyAddress()
-                            : undefined,
-                    senderAddress: cashlinkRequest.senderAddress
-                            ? cashlinkRequest.senderAddress.toUserFriendlyAddress()
+                    appName: createCashlinkRequest.appName,
+                    senderAddress: createCashlinkRequest.senderAddress
+                            ? createCashlinkRequest.senderAddress.toUserFriendlyAddress()
```
sisou: This and similar occurences can be shortened to
```ts
senderAddress: createCashlinkRequest.senderAddress?.toUserFriendlyAddress(),
```
with [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) in Typescript 3.7.

src/lib/RequestParser.ts
```diff
@@ -191,18 +242,29 @@ export class RequestParser {
...
                             : undefined,
-                    senderBalance: cashlinkRequest.senderBalance,
-                } as CashlinkRequest;
+                    senderBalance: createCashlinkRequest.senderBalance,
+                    value: createCashlinkRequest.value,
+                    message: createCashlinkRequest.message,
+                    theme: createCashlinkRequest.theme,
+                    returnCashlink: createCashlinkRequest.returnCashlink,
+                    skipSharing: createCashlinkRequest.skipSharing,
+                } as CreateCashlinkRequest;
```
sisou: Can possibly be shortened with object spreading.

src/views/CashlinkReceive.vue
```diff
@@ -173,6 +191,10 @@ export default class CashlinkReceive extends Vue {
         hubApi.on(HubApi.RequestType.LOGIN, handler, handler);
         hubApi.on(HubApi.RequestType.ONBOARD, handler, handler);
         hubApi.checkRedirectResponse();
+
+        this._onResize = this._onResize.bind(this);
+        window.addEventListener('resize', this._onResize);
+        this._onResize();
```
sisou: I would massively prefer to do breakpoint-display logic with CSS breakpoints instead of with JS.

src/lib/Cashlink.ts
```diff
@@ -57,6 +57,24 @@ class Cashlink {
...
+        if (this._theme && (this._immutable || this.state !== CashlinkState.UNCHARGED)) {
+            throw new Error('Cashlink is immutable');
+        }
+        if (!Object.values(CashlinkTheme).includes(theme) || !Nimiq.NumberUtils.isUint8(theme)) {
+            throw new Error('Unsupported theme');
+        }
+        this._theme = theme;
+    }
+
+    get isThemeEncodedInLink() {
```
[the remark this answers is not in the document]
danimoh: `hasTheme` might suggest that `hasTheme() === true` iff `!!theme`.
However, `!!_theme` is what we're interested in here.
I will use `hasEncodedTheme`.

src/lib/PublicRequestTypes.ts
danimoh: I'll change it to `link` and rename `returnCashlink` to `returnLink` in the request accordingly.

src/lib/RequestParser.ts
danimoh: This also holds true for most other requests.
Can be addressed as part of https://github.com/nimiq/hub/issues/358.

src/views/CashlinkReceive.vue
danimoh: I also generally prefer css media queries and do that if possible.
I this case an `isMobile` property is used to be able to use conditional display via `v-if` in the template to avoid loading both, the desktop and mobile theme, if only one is needed.
`display: none` would not achieve this behavior as images in `` elements are always loaded.
While with css a `background-image` could be applied only when an image is needed, we're using ``s here to enable fade-in after load.

danimoh:
I pushed some changes as https://github.com/nimiq/hub/commit/75404faf0d97caef4f713aec051337bcd88439f0

PR #388, opened 2020-01-24
Update and customize browser warnings

danimoh:
Update to browser warning 1.0.0 which now provides a share button,
fixes issues with old browsers, updated the warning for Edge
browsers, adds types and is minified.
The share button is customized such that it shares the caller url
instead of the hub itself which wouldn't make sense as the hub
request would not be available anymore when opening in another
tab or browser. Cashlinks however are shared normally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
         "postinstall": "cd client && yarn"
     },
     "dependencies": {
-        "@nimiq/browser-warning": "^0.4.0",
+        "@nimiq/browser-warning": "^1.1.0",
         "@nimiq/iqons": "^1.5.0",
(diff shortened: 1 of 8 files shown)


danimoh:
I updated to a new version 1.1.0 with inlined css.
1.0.0 changed the absolute path to `browser-warning.css` to a relative path which is required for apps hosted in sub paths like `nimiq.com/accept-donations`.
The implication or assumption of this is, that `browser-warning.css` needs to be hosted in the same folder as `index.html` with is not the case for the `/cashlink` route.
To avoid this limitation, the css is now inlined in 1.1.0.

PR #393, opened 2020-01-27
Do not allow recipient and sender to be equal in Checkout

### LINKED ISSUE #344: Checkout allows to select same address for sender as recipient
When selecting a Nimiq address to send from in the address selector during checkout, and I have the shop address in my list, it allows me to select it, which leads to a closing of the popup because of a 'recipient and sender addresses cannot match' error.

nibhar:
resolves #344

### DIFF
--- a/src/components/NimiqCheckoutOption.vue
+++ b/src/components/NimiqCheckoutOption.vue
@@ -63,8 +63,8 @@
         </template>
         <template v-else>
             <h2 class="nq-h1">Choose an Address to pay</h2>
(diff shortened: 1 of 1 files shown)


src/components/NimiqCheckoutOption.vue
```diff
@@ -380,8 +404,17 @@ export default class NimiqCheckoutOption
     }
 
     private get hasSufficientBalanceAccount(): boolean {
```
sisou: The function name is now not accurate anymore, since it also tests for eligibility with the recipient address. So can be renamed to `hasEligibleAddress()`.

PR #396, opened 2020-01-28
RpcApi: Replace routerPush and routerReplace with navigation guards

### LINKED ISSUE #371: Replace RpcApi.routerPush and routerReplace by router navigation guards
`routerPush` and `routerReplace` in `RpcApi.ts` can be replaced by [Vue Router Navigation Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards).
`Before Guards` can be used to inject the query and `After Hooks` to export the state.
The query should only be injected when not already set by the initial navigation request as in the call in `_reply`.

nibhar:
This PR removes `RpcApi.routerPush()` and `RpcApi.routerReplace()`. They are replaced with navigation guards on the router. 

As a remark for future use, that does also mean, that future calls to `Vue.$router.push()` now take a path, i.e `Vue.$router.push('/sign-transaction')` as argument. If a named route is used `Vue.$router.push({name: 'sign-transaction'})` must be used. That is different from how `Vue.$rpc.routerPush()` worked before this PR.

I replaced all current occurrences.

resolves #371

### DIFF
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -92,20 +92,20 @@ export default class RpcApi {
         ]);
 
         this._router.beforeEach((to: Route, from: Route, next: (arg?: string | false | Route) => void) => {
-            // // There is an intial redirect from '/' to '/' which does not need to be handled at all.
+            // There is an intial redirect from '/' to '/' which does not need to be handled at all.
             if (to.name === REQUEST_ERROR || (to.path === '/' && from.path === '/')) {
                 next();
                 return;
             }
-            // If the target already has its rpcId, do not update it, since the caller to push()/replace()
-            // can provide it.
+
+            // If the navigation call already contains an rpcId, do not replace it.
             if ( to.query.rpcId ) {
                 next();
                 return;
             }
 
             const rpcId = from.query.rpcId || this._parseUrlParams(window.location.search).rpcId;
-            // In case no rpcId can be found the request is not functional and needs to be rejected.
+            // In case no rpcId can be found, the request is not functional and needs to be rejected.
             if (!rpcId) {
                 this.reject(new Error('UNEXPECTED: RpcId not present'));
                 next(false);
@@ -126,7 +126,7 @@ export default class RpcApi {
             if (to.path === '/' && from.path === '/') {
                 return;
             }
-            // In case there is an rpcState export the entire state to the newly pushed history entry
+            // If we have an rpcState, export the entire state to the newly pushed history entry
             // to be available on reload.
             // This is potentially redundand to the above condition but added as a precaution,
             // especially considering the no-request case a few lines down within RpcApi.start().
--- a/src/lib/RpcApi.ts
+++ b/src/lib/RpcApi.ts
@@ -2,7 +2,7 @@ import { ResponseStatus, RpcServer, State as RpcState } from '@nimiq/rpc';
 import { BrowserDetection } from '@nimiq/utils';
 import { RootState } from '@/store';
 import { Store } from 'vuex';
-import Router from 'vue-router';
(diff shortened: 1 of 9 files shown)


src/lib/RpcApi.ts
```diff
@@ -393,11 +421,11 @@ export default class RpcApi {
 
         if (error.message === KeyguardErrors.Messages.EXPIRED) {
             // Don't reject but navigate to checkout to display the expiration warning there.
-            this.routerReplace(RequestType.CHECKOUT);
+            this._router.replace({ name: RequestType.CHECKOUT });
             return;
         }
 
-        this.routerReplace(keyguardResponseRouter(command, this._staticStore.request!.kind).reject);
+        this._router.replace(keyguardResponseRouter(command, this._staticStore.request!.kind).reject);
```
danimoh: I think this line is not doing what you want (missing `name`).

src/lib/RpcApi.ts
```diff
@@ -103,8 +149,6 @@ export default class RpcApi {
             if (window.opener === null && window.history.length > 1 && !window.location.hash) {
                 location.href = Config.redirectTarget;
             } else {
-                // Not using routerReplace here as that method exports the state which does not exist when there is no
-                // request.
                 this._router.replace(`/${REQUEST_ERROR}`);
```
danimoh: For this call I now get the following Error:
```
RpcApi.ts:218 Uncaught (in promise) TypeError: Cannot read property 'reply' of undefined
    at RpcApi._reply (RpcApi.ts:218)
    at RpcApi.reject (RpcApi.ts:190)
    at RpcApi._router.beforeEach (RpcApi.ts:110)
    at iterator (vue-router.esm.js:2154)
    at step (vue-router.esm.js:1880)
    at runQueue (vue-router.esm.js:1888)
    at HTML5History.confirmTransition (vue-router.esm.js:2181)
    at HTML5History.transitionTo (vue-router.esm.js:2068)
    at HTML5History.replace (vue-router.esm.js:2411)
    at eval (vue-router.esm.js:2824)
```

nibhar:
> In popups I always get this error now:
> 
> ```
> Uncaught (in promise) undefined
> (anonymous) | @ | vue-router.esm.js:2085
>   | abort | @ | vue-router.esm.js:2116
>   | (anonymous) | @ | vue-router.esm.js:2165
>   | RpcApi._router.beforeEach | @ | RpcApi.ts:115
>   | iterator | @ | vue-router.esm.js:2154
>   | step | @ | vue-router.esm.js:1880
>   | runQueue | @ | vue-router.esm.js:1888
>   | confirmTransition | @ | vue-router.esm.js:2181
>   | transitionTo | @ | vue-router.esm.js:2068
>   | replace | @ | vue-router.esm.js:2411
>   | (anonymous) | @ | vue-router.esm.js:2824
>   | replace | @ | vue-router.esm.js:2823
>   | _hubApiHandler | @ | RpcApi.ts:331
>   | async function (async) |   |  
>   | _hubApiHandler | @ | RpcApi.ts:257
>   | _server.onRequest | @ | RpcApi.ts:235
>   | _receive | @ | rpc.es.js:917
> ```

Apparently the vue-router update introduces routing with `Promise<Route>` return values. Every time a routing fails (i.e due to the destination changing, which is what we do) it throws this error. I reverted the upgrade of vue-router, as we do not need it.
If we upgrade in the future we will have to adapt to the promise based api of vue-router.
See https://github.com/vuejs/vue-router/issues/2881#issuecomment-520554378

src/lib/RpcApi.ts
nibhar: Should be fixed now, please take a look.

PR #399, opened 2020-03-04
Checkout fixes

danimoh:
A variety of checkout fixes. See commit history for details.

You can disable white space changes for easier review.

### DIFF
--- a/src/components/CheckoutCard.vue
+++ b/src/components/CheckoutCard.vue
@@ -24,6 +24,7 @@ export default class CheckoutCard<
     protected checkNetworkInterval: number = -1;
     protected paymentState: PaymentState = PaymentState.NOT_FOUND;
     protected selected: boolean = false;
+    protected hasCurrencyInfo!: boolean;
     protected showStatusScreen: boolean = false;
     protected statusScreenState = StatusScreen.State.LOADING;
     protected statusScreenTitle: string = '';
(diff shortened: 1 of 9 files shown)


src/components/CheckoutCard.vue
```diff
@@ -33,6 +34,9 @@ export default class CheckoutCard<
     protected statusScreenMainAction: () => void = () => console.warn('statusScreenMainAction not set.');
 
     protected async created() {
+        this.hasCurrencyInfo = this.request.paymentOptions.length > 1
+            && (!history.state || !history.state[HISTORY_KEY_SELECTED_CURRENCY]);
```
sisou: I don't quite understand the naming of this variable? Why is `hasCurrencyInfo` only true when there are _more_ than one paymentOption? Doesn't the single payment option also include a currency info?

src/views/Checkout.vue
```diff
@@ -154,6 +167,9 @@ class Checkout extends Vue {
         const minWidth = 675; // Width below which disclaimer would break into three lines.
         const minHeight = 890; // Height at which two lines fit at bottom, also if logos over carousel shown.
         this.screenFitsDisclaimer = window.innerWidth >= minWidth && window.innerHeight >= minHeight;
+        // Throttle calls to carousel.updateDimensions as its an expensive call
+        clearTimeout(this.dimensionsUpdateDelay);
+        this.dimensionsUpdateDelay = window.setTimeout(() => (this.$refs.carousel as Carousel).updateDimensions(), 150);
```
sisou: As a "delay" is the time (150ms in this case), a better name for this variable would be either `dimensionsUpdateThrottle` or `dimensionsUpdateTimeout`, I think.
```suggestion
        this.dimensionsUpdateThrottle = window.setTimeout(() => (this.$refs.carousel as Carousel).updateDimensions(), 150);
```

src/components/CheckoutCard.vue
danimoh: Yes, exactly. This has always been the case.

PR #407, opened 2020-03-29
CheckoutServerApi: serve dummy data in development mode

danimoh:
This PR proposes to add checkout dummy data similar to the snippet already in use by Matheo, Sebastian and me to actual code base in a safe way.
This removes the need to always apply the snippet as a patch or from a stash when needed and the extra care to not include it in commits or builds. Additionally, it makes testing / developing easier for third party devs, that would need to write their own snippet.

Might be sufficient to resolve https://github.com/nimiq/hub/issues/377.

### DIFF
--- a/src/lib/CheckoutServerApi.ts
+++ b/src/lib/CheckoutServerApi.ts
@@ -130,41 +130,54 @@ export default class CheckoutServerApi {
     }
 
     private static _generateDummyData(requestData: URLSearchParams) {
-        const now = Date.now();
-        const options = [
-            {
-                currency: Currency.BTC,
-                type: PaymentType.DIRECT,
-                amount: '.0002e8',
-                expires: now + 15 * 60000, // 15 minutes
-                protocolSpecific: {
-                    feePerByte: 2, // 2 sat per byte
-                    recipient: '17w6ar5SqXFGr786WjGHB8xyu48eujHaBe', // Unicef
+        if (requestData.get('command') === 'get_state') {
+            return {
+                time: Date.now(),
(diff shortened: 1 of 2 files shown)


src/lib/CheckoutServerApi.ts
```diff
@@ -112,6 +112,12 @@ export default class CheckoutServerApi {
         }
 
         if (!fetchResponse.ok) {
+            if (process.env.NODE_ENV === 'development'
```
sisou: I don't think this place is a good way to put the check, as with this in development the dev will not notice a failing callback url, making testing of error codes or even the callback at all impossible.

I'd say it should only fallback to dummy data if explicitly requested, e.g. by using a constant `callbackUrl` in the request to trigger dummy data.

danimoh: Who do you mean by dev?
For a store dev using a pre-built hub client this line will result in false.

sisou: Hm, yes you are right. For deployed versions this is `false`. Ok, I guess it's fine then.

Have you tested this? Because I believe in the case where the provided `callbackUrl` is not valid, or rather does not respond at all, `fetch` throws an error both times, which then means that `fetchResponse` will be `undefined` and you'll get an error of  `'Cannot read property ok of undefined'` in the if-condition.

sisou: I misunderstood. When the second fetch throws, execution stops and we won't even get to line 114. :facepalm:

PR #409, opened 2020-04-05
Add Easter theme

danimoh:


src/views/CashlinkReceive.vue
```diff
@@ -341,6 +342,7 @@ class CashlinkReceive extends Vue {
             case CashlinkTheme.GENERIC:
             case CashlinkTheme.CHRISTMAS:
             case CashlinkTheme.LUNAR_NEW_YEAR:
+            case CashlinkTheme.EASTER:
                 return 'You are loved';
```
Yukiioz, line 346: maybe "You are loved" for Christmas & Lunar new year. But "Happy easter" or something like that for easter, nah ?

src/lib/Cashlink.ts
```diff
@@ -480,15 +480,9 @@ namespace Cashlink {
...
 
     // To be updated with the seasons.
-    // Currently enable LUNAR_NEW_YEAR theme for users that speak Chinese, Korean, Vietnamese, Malay or Indonesian (note
-    // that language tags follow BCP47 which starts with a ISO639 language and optionally ends with a ISO3166 country).
-    // Disable after the Lantern Festival, which ends Feb 8th.
-    export const DEFAULT_THEME = Date.now() < new Date('Wed, 9 Feb 2020 23:59:00 GMT+8').valueOf()
-        && navigator.languages.find(
-            (lang) => /^(zh|cmn|yue|gan|ko|vi|may|msa|mal|id)\b|-(CN|HK|TW|MO|KP|KR|VN|MY|SG|ID)$/i.test(lang))
-        ? CashlinkTheme.LUNAR_NEW_YEAR
+    export const DEFAULT_THEME = Date.now() < new Date('Tue, 13 Apr 2020 23:59:00 GMT-12').valueOf()
```
sisou, line 483: Maybe this can be generalized, so we can give a date range to a theme so it happens every year, without having to update the code every couple of weeks?

danimoh: I thought similarly, but let's do it next time when actually an event actually reoccurs for the first time.

PR #413, opened 2020-04-19
Add vendorMarkup option and checkout price tooltip

danimoh:
Requires https://github.com/nimiq/vue-components/pull/81.

### DIFF
--- a/src/lib/CheckoutServerApi.ts
+++ b/src/lib/CheckoutServerApi.ts
@@ -141,6 +141,7 @@ export default class CheckoutServerApi {
                     currency: Currency.BTC,
                     type: PaymentType.DIRECT,
                     amount: '.0002e8',
+                    vendorMarkup: -.05,
                     expires: now + 15 * 60000, // 15 minutes
                     protocolSpecific: {
                         feePerByte: 2, // 2 sat per byte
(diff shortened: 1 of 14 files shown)


src/lib/CheckoutServerApi.ts
```diff
@@ -141,6 +141,7 @@ export default class CheckoutServerApi {
                     currency: Currency.BTC,
                     type: PaymentType.DIRECT,
                     amount: '.0002e8',
+                    vendorMarkup: -.05,
```
sisou, line 144: Is this a different amount than in Demo.ts on purpose? And in the ETH option and the gasPrice?

src/lib/PublicRequestTypes.ts
```diff
@@ -102,6 +102,10 @@ export interface PaymentOptions<C extends Currency, T extends PaymentType> {
      * i.e Luna for Currency.NIM and Satoshi for Currency.BTC
      */
     amount: string;
+    /**
+     * Crypto payment markup contained in `amount` that the vendor charges. Relative value, e.g. .01 for 1%.
+     */
+    vendorMarkup?: number;
```
sisou, line 108:
```suggestion
    /**
     * Crypto payment markup contained in `amount` that the vendor charges. Relative value, e.g. .01 for 1%.
     * Can be negative to describe a discount.
     */
    vendorMarkup?: number;
```

src/lib/CheckoutServerApi.ts
danimoh: Yes, to test the update functionality.

danimoh:
Will keep this as a draft until there's also a new Keyguard client with support for the vendor markup.

PR #414, opened 2020-04-19
Some checkout adaptions

danimoh:
- Checkout disclaimer update
- Disable disclaimer for voting app
- Fix hubApi.checkout return type

As draft as the url for the voting app is potentially not final yet.

### DIFF
--- a/client/HubApi.ts
+++ b/client/HubApi.ts
@@ -124,11 +124,16 @@ export default class HubApi<DB extends BehaviorType = BehaviorType.POPUP> { // D
         return this._request(requestBehavior, RequestType.MANAGE_CASHLINK, [request]);
     }
 
-    public checkout<B extends BehaviorType = DB>(
-        request: Promise<CheckoutRequest> | CheckoutRequest,
+    public checkout<R extends CheckoutRequest, B extends BehaviorType = DB>(
+        request: Promise<R> | R,
         requestBehavior: RequestBehavior<B> = this._checkoutDefaultBehavior as any,
-    ): Promise<B extends BehaviorType.REDIRECT ? void : SignedTransaction> {
-        return this._request(requestBehavior, RequestType.CHECKOUT, [request]);
+    ): Promise<B extends BehaviorType.REDIRECT
+        ? void
+        : R extends { version: 2 }
+            ? SimpleResult | SignedTransaction
+            : SignedTransaction
+    > {
+        return this._request(requestBehavior, RequestType.CHECKOUT, [request]) as any;
     }
 
     public chooseAddress<B extends BehaviorType = DB>(
--- a/src/views/Checkout.vue
+++ b/src/views/Checkout.vue
@@ -62,8 +62,9 @@
                 @close="_closeDisclaimerOverlay"
             >
                 <strong>Disclaimer</strong>
-                This Nimiq interface is non-custodial and solely used to bridge the customer with the merchant directly
-                (P2P). Payment and order fulfillment are sole responsibility of the customer and merchant respectively.
+                This Nimiq interface is non-custodial and solely used to bridge the payment sender with the payment
+                recipient directly (P2P). Payment and deliverable fulfillment for payment are sole responsibility of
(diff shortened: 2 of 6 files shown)


src/views/Checkout.vue
```diff
@@ -103,6 +104,8 @@ class Checkout extends Vue {
     private initialCurrencies: PublicCurrency[] = [];
     private availableCurrencies: PublicCurrency[] = [];
     private readonly isIOS: boolean = BrowserDetection.isIOS();
+    private readonly disclaimerDisabled: boolean = // disable disclaimer for voting app
+        /^https?:\/\/(www\.)?nimiq(-testnet)?\.com\/vote\b/.test(document.referrer);
```
sisou, line 108: Seems like this would be a good case for a config setting again, like with the NIM-less checkout origins.

A referrer might not be available in all browsers though, as some browsers or browser extensions might block all referrers, or block the `path` part of a referrer.

danimoh: In the config we only have options that depend on the origin so far, therefore I was hesitant to add this option (which depends on a path) to the config to avoid having to come up with a complicated config name.
I think having a regex check here is good enough.

Do you have more insights about blocked referrers / paths?
I tested in Chrome and Firefox and I was able to read the referrer including the path.
An alternative would be to add a `disableDisclaimer` option for privileged origins to the Checkout request, further complicating that request.

danimoh:
I changed to the final url of the voting app and removed the draft status.

src/views/Checkout.vue
sisou: https://chrome.google.com/webstore/detail/referer-control/hnkcfpcejkafcihlgbojoidoihckciin for example is an extension for Chrome that lets people disable the referrer globally.

danimoh: Well, such extensions do not seem to be wide spread and extensions can always break things.
But I changed it nonetheless to a `disableDisclaimer` flag. In the end, I think its a little bit nicer.

danimoh:
I switched to a `disableDisclaimer` flag and added additional improvements to the checkout return type.

client/HubApi.ts
```diff
@@ -124,11 +124,16 @@ export default class HubApi<DB extends BehaviorType = BehaviorType.POPUP> { // D
...
+    public checkout<R extends CheckoutRequest, B extends BehaviorType = DB>(
+        request: Promise<R> | R,
         requestBehavior: RequestBehavior<B> = this._checkoutDefaultBehavior as any,
-    ): Promise<B extends BehaviorType.REDIRECT ? void : SignedTransaction> {
-        return this._request(requestBehavior, RequestType.CHECKOUT, [request]);
+    ): Promise<B extends BehaviorType.REDIRECT
+        ? void
+        : R extends { version: 2 }
+            ? SimpleResult | SignedTransaction
+            : SignedTransaction // Note: this will not be guaranteed anymore when checkout supports external nim wallets
```
sisou, line 134: Since we still have the Nimiq Network and have direct access to the blockchain to monitor for external tx in the browser, we _do still get_ the signed tx from the network and can simply return it.

So I think this return type will not change.

client/HubApi.ts
```diff
@@ -124,11 +124,16 @@ export default class HubApi<DB extends BehaviorType = BehaviorType.POPUP> { // D
...
+    > {
+        return this._request(requestBehavior, RequestType.CHECKOUT, [request]) as any;
```
sisou, line 136: Is `as any` necessary now or is it a leftover from experimenting?

client/HubApi.ts
danimoh: Yes, could be

client/HubApi.ts
danimoh: It's required unfortunately because `ResultByRequestType` distringuishes the result only by request type, not request, and therefore gives `SimpleResult | SignedTransaction`. But `SimpleResult` can not be mapped to `SignedTransaction`.

PR #420, opened 2020-06-12
Fix transaction sending in case of pico->nano upgrade

sisou:
In case of pico -> nano consensus upgrade, the `transaction-relayed` listener on the consensus is lost, which means `Network.sendToNetwork(tx)` never resolves, showing the loading screen forever.

This PR fixes that, similar to how it was fixed in the Safe. Additionally, it adds error handling in case the transaction fails to send (with the options to reload the page or cancel the request).

@danimoh I don't know how the `SignTransactionLedger` component can support a tx-sending error, so I have not implemented that. Please have a look if we can make that possible.

**This branch is now deployed in the testnet.**

### DIFF
--- a/src/views/CheckoutTransmission.vue
+++ b/src/views/CheckoutTransmission.vue
@@ -1,7 +1,17 @@
 <template>
     <div class="container pad-bottom">
         <SmallPage>
-            <StatusScreen :title="title" :status="status" :state="state" :lightBlue="true"/>
+            <StatusScreen
+                :title="title"
+                :status="status"
+                :state="state"
+                :message="message"
+                mainAction="Reload"
+                alternativeAction="Cancel"
+                @main-action="reload"
+                @alternative-action="cancel"
+                lightBlue
+            />
             <Network ref="network" :visible="false"/>
         </SmallPage>
     </div>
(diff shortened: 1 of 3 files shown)


src/views/CheckoutTransmission.vue
```diff
@@ -1,7 +1,17 @@
...
+            <StatusScreen
+                :title="title"
+                :status="status"
+                :state="state"
+                :message="message"
+                mainAction="Reload"
+                alternativeAction="Cancel"
+                @main-action="reload"
+                @alternative-action="cancel"
+                :lightBlue="true"
```
danimoh, line 13: Might want to just use `lightBlue`, same as in `CashlinkManage`.

src/views/CheckoutTransmission.vue
```diff
@@ -34,10 +46,17 @@ export default class CheckoutTransmission extends Vue {
...
-        setTimeout(() => this.$rpc.resolve(result), StatusScreen.SUCCESS_REDIRECT_DELAY);
+        try {
+            const result = await (this.$refs.network as Network).sendToNetwork(tx);
+            this.state = StatusScreen.State.SUCCESS;
+
+            setTimeout(() => this.$rpc.resolve(result), StatusScreen.SUCCESS_REDIRECT_DELAY);
+        } catch (error) {
+            this.state = StatusScreen.State.WARNING;
+            this.message = error.message;
+            return;
```
danimoh, line 58: This return is not necessary.

PR #431, opened 2020-11-18
Swap

sisou:
Currently based on `bitcoin`, will be re-based onto `master` when #430 is merged.


Yukiioz (approved):
Well done, didn't found any issue worth mentioning.
Good job.

PR #436, opened 2021-02-14
Generate UIDs for WalletInfos and return them to the Wallet

sisou:
This PR adds a new field to the `Account` type returned to the Wallet: a `uid`. It is returned everywhere where an account is returned.
Iframes in iOS/Safari, which don't have the `keyId` stored in the cookie, return an empty string. The wallet falls back to tracking swap volume via transaction history when the UID is not available, but since it is returned for signups, logins, renames, exports, it will soon be available for those users also.

Generating the UID is a runtime SHA256 hash, which is super fast and not noticeable until ~100 accounts per browser.

> The UID is used for the purpose of tracking Fastspot swap limits per user. It is generated from two deterministic values. The keyId of an account and its first NIM address, wich are always the same.
>
> The `keyId` is never passed to outside the Hub, so can be seen as a secret value. This way it is impossible for anyone who gets access to the UID alone to determine the user's address.

### DIFF
--- a/src/lib/WalletInfo.ts
+++ b/src/lib/WalletInfo.ts
@@ -10,6 +10,8 @@ import { labelKeyguardAccount } from './LabelingMachine';
 import WalletInfoCollector from './WalletInfoCollector';
 import { WalletStore } from '../lib/WalletStore';
 import { WalletType } from './Constants';
+import { makeUid } from './Uid';
+import AddressUtils from './AddressUtils';
 
 export class WalletInfo {
     public static fromObject(o: WalletInfoEntry): WalletInfo {
@@ -33,9 +35,11 @@ export class WalletInfo {
             o.keyMissing, o.fileExported, o.wordsExported, o.btcXPub, btcAddresses);
     }
 
-    public static objectToAccountType(o: WalletInfoEntry): Account {
+    public static async objectToAccountType(o: WalletInfoEntry): Promise<Account> {
         // Polyfill BTC address lists for pre-BTC wallets
         if (!o.btcAddresses) o.btcAddresses = { internal: [], external: [] };
+
+        const accountInfoEntries = Array.from(o.accounts.values());
 
         return {
             accountId: o.id,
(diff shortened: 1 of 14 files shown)


src/lib/WalletInfo.ts
```diff
@@ -201,8 +210,14 @@ export class WalletInfo {
                 internal: this.btcAddresses.internal.map((btcAddressInfo) => btcAddressInfo.toBtcAddressType()),
                 external: this.btcAddresses.external.map((btcAddressInfo) => btcAddressInfo.toBtcAddressType()),
             },
+            uid: await this.getUid(),
```
danimoh, line 213: The way the `uid` is generated and returned is different from how it's done for other properties like the account id.
Previously, `toAccountType` was just converting the data format of `WalletInfo` instances without adding or generating data.
All other data is previously generated and added to the wallet instance and also stored in the wallet store (also taking up a bit of storage space obviously).
So question is which approach we prefer.

Maybe @Yukiioz also has an opinion on this.

Yukiioz (approved):
Found nothing to comment. Except maybe the `makeUid` method that could be renamed into something that match more other functions in the Hub (i don't think we have any other make... functions). But that's really a minor thing.

Nicely done. 👍

danimoh (changes requested):
Note that for Ledger accounts the keyId is calculated from the first addresses' public key.
Therefore for Ledgers, the uid can be calculated by only knowing that public key.
If the user actively uses this address, external people can DOS this user by eating up his limits by setting up unused swaps.
To mitigate this I'd propose to use an address at an obscure key path that will not be used on the blockchain by the user instead of the first address for the uid calculation.

src/lib/WalletInfo.ts
sisou: The difference is that generating the UID is async, and all other data generation is sync. I would like to generate the UID also in the constructor already, but constructors cannot be async.

danimoh: For the account id, we pass the asynchronously pre-derived id into the constructor.

danimoh (approved):

danimoh:
We decided to keep the behavior as is currently, and potentially change the Ledger WalletID calculation in the ledger api soon after (with migration to the new id in the hub).

PR #445, opened 2021-05-10
Add SRI

paberr:
Let Vue.js automatically add SRI to all script tags it produces.
There are two scripts in the `index.html` that still require the SRI to be manually set:
`browser-warning.js` and the Core JS library.

What would be the best way to set these? Should it be hardcoded into the html or somehow be part of the build process?

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
         "typescript": "~3.8.0",
         "vue-template-compiler": "^2.6.11",
         "webpack-i18n-tools": "https://github.com/nimiq/webpack-i18n-tools#master",
+        "webpack-subresource-integrity": "^1.5.2",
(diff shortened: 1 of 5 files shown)


Yukiioz (approved):
Thank you for the PR again @paberr ! Nice and simple :) 

To answer your question, i think it would be better to have it included in the build process for sure. But since those scripts won't change very often we could simply hardcore it if you prefer.

Yukiioz:
So i tried to find a way to generate sri hash for browser-warning.js and the Core JS library during the build process, but the simple solutions did not seem to work. It means that for now, those are hardcoded until i spend more time to automate it.

danimoh:
See https://github.com/nimiq/hub/pull/449 for follow-up PR.

PR #446, opened 2021-05-15
windows fix

neokrept:


danimoh (approved):
Yep, this still works on Linux too :+1:

Just note that our branch names are usually prefixed by the author name, so this could have been `sabin/build-fix-draft`.

PR #453, opened 2021-06-27
Temporarily circumvent Desktop Safari resource integrity check bug client-side

cb5890:
ServiceWorker.js: drop all headers from request that could trigger a HTTP 304 response

This commit contains a temporary workaround to deal with a bug in Desktop Safari
that prevents a successful resource integrity check when
  a resource is loaded via the service worker
  *and* the webpage forces cache revalidation (like we do).

When the resource is not modified, the server responds with a HTTP "304 -
Not modified" response. However, then Safari (supposedly) tries to check
the content from *this* response against the integrity hash *instead* of
the actual cached content. Obviously, this is pointless, since a HTTP
304 doesn't contain any content, because that would defer the entire
point, right? Hence, the integrity check won't ever succeed.. :-/

Therefore, we're now removing all headers from the request that could trigger
such a response. Addmitedly, that's less then ideal, since it effectiviely breaks
caching for all resources with an integrity check. However, these resources are
typically small and this patch is *intended* to be temporary.

### DIFF
--- a/public/ServiceWorker.js
+++ b/public/ServiceWorker.js
@@ -18,6 +18,22 @@ self.addEventListener('fetch', event => {
     // See: https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#Cross-origin_data_storage_access
     // @ts-ignore Property 'request' does not exist on type 'Event'.ts
     const requestHost = new URL(event.request.url).host;
+
+    let headers = event.request.headers;
+    // Check whether a resource specifies an integrity hash.
+    if (event.request.integrity != '') {
+        // Create a duplicate of the request headers, since these are read-only now.
+        headers = new Headers();
+        for (const [key, value] of event.request.headers.entries()) {
+            // Drop all headers that could trigger a "304 - Not modified" HTTP response.
+            // This helps to prevent Desktop Safari from being unable to verify the
+            // integrity of the fetched ressource due to a seemingly empty response... :-/
+            if (key != 'if-modified-since' && key != 'if-none-match') {
+                headers.set(key, value);
+            }
+        }
+    }
+
     if (requestHost === location.host) {
         // forward request
         // @ts-ignore Property 'respondWith' does not exist on type 'Event'.ts,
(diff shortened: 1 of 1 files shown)


public/ServiceWorker.js
```diff
@@ -18,13 +18,31 @@ self.addEventListener('fetch', event => {
...
+        // Create a duplicate of the request headers, since these are read-only now.
+        headers = new Headers();
+        for (const [key, value] of event.request.headers.entries()) {
+            // Drop all headers that could trigger a "304 - Not modified" HTTP response.
+            // This helps to prevent Desktop Safari from being unable to verify the
+            // integrity of the fetched ressource due to a seemingly empty response... :-/
+            if (key != 'if-modified-since' && key != 'if-none-match') {
+                headers.set(key, value);
+            }
+        }
```
danimoh, line 34:
```suggestion
        headers = new Headers(event.request.headers);
        // Drop all headers that could trigger a "304 - Not modified" HTTP response.
        // This helps to prevent Desktop Safari from being unable to verify the
        // integrity of the fetched resource due to a seemingly empty response... :-/
        // Note that currently, the ServiceWorker is only installed for Safari, therefore
        // we don't need to check for the browser here to apply this fix selectively for
        // Safari only.
        // TODO: check in the future whether Safari still needs this special treatment
        headers.delete('If-Match');
        headers.delete('If-None-Match');
        headers.delete('If-Modified-Since');
        headers.delete('If-Unmodified-Since');
```
Slightly more elegant like this imo.
Additionally, filtering some more headers listed on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#conditionals.

public/ServiceWorker.js
```diff
@@ -18,13 +18,31 @@ self.addEventListener('fetch', event => {
...
+
     if (requestHost === location.host) {
         // forward request
         // @ts-ignore Property 'respondWith' does not exist on type 'Event'.ts,
         // Property 'request' does not exist on type 'Event'.ts
         event.respondWith(fetch(event.request, {
             // omit cookie transmission
             credentials: 'omit',
+            // override headers as they've might changed
+            headers: headers
```
danimoh, line 45:
```suggestion
            headers,
```
Simplified syntax and adding trailing comma.

danimoh (approved):
Thanks for investigating this and the fix :)

I added some suggestions (which I didn't test).
Wdyt?

cb5890:
I've just tested it again and it works as intended. Ready to be deployed :)

PR #509, opened 2023-06-26
Allow returning USDC address from choose-address request

sisou:
And return account info, too.

This info is used by CPL first.


client/PublicRequestTypes.ts
```diff
@@ -652,7 +660,7 @@ export type ResultByRequestType<T> =
     T extends RequestType.ONBOARD | RequestType.SIGNUP | RequestType.LOGIN
             | RequestType.MIGRATE | RequestType.LIST ? Account[] :
     T extends RequestType.LIST_CASHLINKS ? Cashlink[] :
-    T extends RequestType.CHOOSE_ADDRESS | RequestType.ADD_ADDRESS ? Address :
+    T extends RequestType.CHOOSE_ADDRESS | RequestType.ADD_ADDRESS ? ChooseAddressResult :
```
danimoh, line 663: `RequestType.ADD_ADDRESS` should still just return an `Address`.

danimoh (approved):

client/PublicRequestTypes.ts
sisou: oh, you are right! Overlooked this.

PR #526, opened 2024-02-21
Fix typos

rex4539:


rex4539:
@sisou Could you please take a look?

sisou (changes requested):
Thanks for this. Please revert the "fix" in `BitcoinJS.min.js`. That is a generated file from a dependency.

sisou (approved):

PR #549, opened 2024-11-01
Add support for swapping USDT

sisou:


sisou:
Test fails because of a Typescript error with the new ledger-api package.

PR #550, opened 2024-11-13
Change API to use a number array instead of a Uint8Array

blouflashdb:
```
const clientObj: Uint8Array = new Uint8Array([21, 31]);
const message = JSON.stringify(clientObj)
console.log("clientObj", clientObj instanceof Uint8Array) 

const serverObj: Uint8Array = JSON.parse(message);
console.log("serverObj", serverObj instanceof Uint8Array) // returns false because Uint8Array is serialized as an object
```

With this change SignedMessage can be serialized and deserialized without problems. Its a breaking API change tho

### DIFF
--- a/client/PublicRequestTypes.ts
+++ b/client/PublicRequestTypes.ts
@@ -456,8 +456,8 @@ export interface SignMessageRequest extends BasicRequest {
 
 export interface SignedMessage {
     signer: string; // Userfriendly address
-    signerPublicKey: Uint8Array;
-    signature: Uint8Array;
+    signerPublicKey: number[];
+    signature: number[];
 }
 
 export interface Address {
--- a/docs/api-methods/30_sign-message.md
+++ b/docs/api-methods/30_sign-message.md
@@ -75,8 +75,8 @@ The method returns a `SignedMessage` object containing the following properties:
 ```javascript
 interface SignedMessage {
     signer: string;               // Userfriendly address
-    signerPublicKey: Uint8Array;  // The public key of the signer
-    signature: Uint8Array;        // Signature for the message
+    signerPublicKey: number[];  // The public key of the signer
+    signature: number[];        // Signature for the message
 }
 ```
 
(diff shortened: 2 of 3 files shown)


sisou:
Just like you are doing `Array.from()` now internally, can't you just do that in your code?

```js
JSON.parse(Array.from(message));
```

Uint8Array cannot be JSON serialized without helpers.

danimoh:
Hi @blouflashdb,
thanks for your considerations and the pull request. I see where you're coming from, but personally I think that `Uint8Array` is in fact more intuitive and consistent. `Uint8Array` makes it very clear that bytes are being returned here, `number[]` could also include `NaN`, negative, fractional or infinite values or values which don't fit a single byte, i.e. the documentation would need to clearly state what kind of values are to be expected, or users of the API strictly speaking would need to check for unsupported values, which ends up being more code even than simply converting the `Uint8Array` to `number[]` as a user. Also, `number[]` is not used anywhere else in the API and keeping `Uint8Array` would therefore be more consistent. I think converting it yourself as a user, as you're currently doing it, is the way to go.

I would favor keeping it a `Uint8Array`.

docs/api-methods/30_sign-message.md
```diff
@@ -75,8 +75,8 @@ The method returns a `SignedMessage` object containing the following properties:
 ```javascript
 interface SignedMessage {
     signer: string;               // Userfriendly address
-    signerPublicKey: Uint8Array;  // The public key of the signer
-    signature: Uint8Array;        // Signature for the message
+    signerPublicKey: number[];  // The public key of the signer
+    signature: number[];        // Signature for the message
```
danimoh, line 79: Should state which values are to be expected here, i.e. integers in the range 0..255.

client/PublicRequestTypes.ts
```diff
@@ -456,8 +456,8 @@ export interface SignMessageRequest extends BasicRequest {
 
 export interface SignedMessage {
     signer: string; // Userfriendly address
-    signerPublicKey: Uint8Array;
-    signature: Uint8Array;
+    signerPublicKey: number[];
+    signature: number[];
```
danimoh, line 460: Should state which values are to be expected here, i.e. integers in the range 0..255.

danimoh (changes requested):
If we were to make this change, some the expected values should be documented, but I think we should keep it being `Uint8Array`s.

PR #553, opened 2024-11-27
Add deploy script, like the one in the Wallet

mraveux:


deploy.sh
```diff
@@ -0,0 +1,403 @@
+#!/bin/bash
+
+# Exit on error
+set -e
```
danimoh, line 4: @hrxi  says we should do this.
He also says the last line might not work on ~~trash~~ Mac.

```suggestion
set -eu
set -o pipefail
```

danimoh (approved):
One suggestion by @hrxi (also applies to the other repositories). Other than that good to go.

PR #554, opened 2024-11-27
Fix max-len warning in WalletInfoCollector.ts

mraveux:


danimoh:
I think this one should not be merged / not yet.
Rather, if our updates make syncing sufficiently quick again, I would revert the change that commented out the original code, and double-commented these lines, causing them to get too long.

PR #562, opened 2024-12-20
Feature: Ledger Wallet Support for Signing Messages

blouflashdb:
# Feature: Ledger Wallet Support for Signing Messages

This pull request introduces the ability to use Ledger Wallets to sign messages.

### 1. StatusMessage Component
- I was unable to get the `StatusMessage` component to work as expected.
- The issue seems to be related to CSS, which I have limited experience with.
- To avoid blocking this feature, I decided to exclude the `StatusMessage` from this PR.
- If needed, I’m happy to provide the groundwork for it, and someone more experienced with CSS can finalize the styling and integration.

### 2. Tab Space UI Feature
- This PR does not include the Tab Space UI feature.
- If this feature is deemed necessary, I suggest someone submits an issue to prioritize and address it in a separate iteration.

## Focus
The primary goal of this PR is to enable message signing with Ledger Wallets, which I believe is the most critical functionality at this time.

Please review and let me know if there are additional adjustments needed!

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4471,9 +4471,9 @@ caniuse-api@^3.0.0:
     lodash.uniq "^4.5.0"
 
 caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001726:
-  version "1.0.30001726"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001726.tgz#a15bd87d5a4bf01f6b6f70ae7c97fdfd28b5ae47"
-  integrity sha512-VQAUIUzBiZ/UnlM28fSp2CRF3ivUn1BWEvxMcVTNwpw91Py1pGbPIyIKtd+tzct9C3ouceCVdGAXxZOpZAsgdw==
+  version "1.0.30001741"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001741.tgz"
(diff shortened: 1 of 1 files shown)


src/views/SignMessageLedger.vue
```diff
@@ -0,0 +1,144 @@
...
+    width: 100%;
+    padding: 1.5rem;
+    flex-grow: 1;
+    resize: none;
+    font-family: 'Fira Mono', monospace;
+}
+
+.bottom-container {
+    height: 23rem;
+}
```
danimoh, line 143: Some styling issues regarding sizes and paddings..

src/views/SignMessageLedger.vue
```diff
@@ -0,0 +1,144 @@
+<template>
+    <div class="container">
+        <SmallPage id="sign-message">
+            <PageHeader>
+                {{ $t('Sign Message') }}
+            </PageHeader>
+
+            <PageBody>
+                <textarea id="message" readonly="readonly" :value="request.message"></textarea>
```
danimoh, line 9: Would be theoretically better to use a `class` instead of an `id` here, as if there were to be a second `SignMessageLedger` instance on the page, it would be a duplicate identifier. Not that it's an actual issue here though, as there won't be two `SignMessageLedger` instances at the same time :)

danimoh (changes requested):
Hello @blouflashdb,

thank you very much for this PR. The functionality is already mostly there :fire: 
However, there are some things that need to be addressed. I will take it from here and work on these changes in the new year.

I wish you a happy new year :champagne:

Albermonte:
> > If you use yarn (classic), it should ignore the nodejs requirement, see `.yarnrc`. Alternatively run `yarn --ignore-engines`, then it'll also work with current Node 22. No need to panic ;)
> 
> Does not seem to work with the latest yarn version (Yarn 4.10.3) and seems like there is no alternative.

https://classic.yarnpkg.com/lang/en/

PR #563, opened 2025-01-15
Redirect to onboard when the user has no wallets

DovAzencot:
Redirect to onboard when the user has no wallets connected and want to sign a message instead of just showing empty list

### DIFF
--- a/src/views/OnboardingSelector.vue
+++ b/src/views/OnboardingSelector.vue
@@ -16,6 +16,7 @@
 
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator';
+import { Getter } from 'vuex-class';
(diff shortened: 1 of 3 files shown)


src/views/OnboardingSelector.vue
```diff
@@ -111,7 +114,9 @@ export default class OnboardingSelector extends Vue {
     }
 
     private close() {
-        if (this.isSecondaryOnboarding) {
+        if (this.isSecondaryOnboarding
+            // In the case we don't have wallets while signing a message, we close the hub
+            && (this.originalRouteName !== RequestType.SIGN_MESSAGE || this.hasWallets)) {
```
sisou, line 119: I don't understand this condition and the comment. The comment talks about _not_ having wallets, but the check is for _having_ wallets.

PR #577, opened 2026-05-05
Update client dependencies and build tooling

Albermonte:
## Summary
- Update `/client` runtime dependencies, build tooling, and lockfile for the dependency refresh.
- Remove direct `/client` usage of `@opengsn/common` by defining the needed GSN v2 `RelayRequest` shape locally.
- Regenerate the tracked `client/dist` bundles and type declarations after the dependency and type changes.

## Testing
- `yarn lint`

## Audit notes
- `/client` no longer installs `@opengsn/common`, so the previous client-side GSN audit exception is gone.
- The root app still receives `@opengsn/common` transitively from `@nimiq/keyguard-client@1.10.0`; those audit findings remain outside issue #576 scope.


sisou:
Where is the opengsn update exception documented?

Albermonte:
> Where is the opengsn update exception documented?

it's a draft, I'm still on it

I was thinking about maybe removing the opengsn dep completely since it's only used for a type and it won't be updated anymore and hardcoding that `RelayRequest` type

Wdty?

sisou:
> I was thinking about maybe removing the opengsn dep completely since it's only used for a type and it won't be updated anymore and hardcoding that `RelayRequest` type
> 
> Wdty?

Yes, that's a good idea!

types/SymbolDispose.d.ts
sisou, line 1: What is this file used for, why is it necessary?

Albermonte: `hub` uses TypeScript 3.8.3
`@nimiq/core` uses TypeScript 5.x 

Typescript introduced in [5.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management:~:text=This%20starts%20by%20adding%20a%20new%20built%2Din%20symbol%20called%20Symbol.dispose%2C%20and%20we%20can%20create%20objects%20with%20methods%20named%20by%20Symbol.dispose) this `Symbol.dispose` in the [web.d.ts generated file](https://cdn.jsdelivr.net/npm/@nimiq/core@2.4.0/types/wasm/web.d.ts#:~:text=Address%20%7B%0A%20%20%20%20free()%3A%20void%3B-,%5BSymbol.dispose%5D()%3A%20void%3B,-__getClassname()%3A%20string%3B%0A%20%20%20%20/**%0A%20%20%20%20%20*%20Compares)

So it's only there to prevent type errors

sisou: So it's for compat with @nimiq/core? When can this file be removed?

Can you add a comment to the file explaining this?

Albermonte: > So it's for compat with @nimiq/core? 

Yes, if not, this will show on console when running `yarn pr`
```sh
ERROR in /Users/albermonte/nimiq/hub/client/node_modules/@nimiq/core/types/wasm/bundler.d.ts(2166,13):
2166:13 Property 'dispose' does not exist on type 'SymbolConstructor'.
    2164 | export class Transaction {
    2165 |     free(): void;
  > 2166 |     [Symbol.dispose](): void;
 ```

> When can this file be removed?

If we update everything to TS 5.2+

sisou (approved):
LGTM 🎉

PR #582, opened 2026-07-13
Fix Ledger confirm-address screen rendering

### LINKED ISSUE #581: LedgerUi: confirm-address screen SVG data URI is missing its leading "<", screen renders blank
In `src/components/LedgerUi.vue`, the `.ledger-screen-confirm-address` rule's background-image data URI begins with `data:image/svg+xml,svg xmlns=...`. The leading `<` of `<svg` is missing, so the browser cannot parse the SVG and drops the background. The confirm-address device screen renders as an empty black screen while every other screen state draws correctly.

All 11 other screen data URIs in the file start with `data:image/svg+xml,<svg`; this is the only one missing the bracket. One-character fix:

```diff
 .ledger-screen-confirm-address {
-    background-image: url('data:image/svg+xml,svg xmlns="http://www.w3.org/2000/svg" fill="white" viewBox="0 0 114 37.5">...');
+    background-image: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" fill="white" viewBox="0 0 114 37.5">...');
 }
```

Location: line 528 at current master (3cee2ef). Found while porting LedgerUi into a pixel-verified component registry: a headless Chromium render at cf2ea41 showed the blank screen, and the malformed URI reproduces at the master tip.

Albermonte:
## What

Restore the missing `<` in the confirm-address SVG data URI.

## Why

The malformed URI cannot be parsed as SVG, leaving the Ledger confirm-address screen blank.

## Validation

- `yarn lint`
- `yarn tsc --noEmit --skipLibCheck`
- `yarn test:unit --runInBand --no-watchman` — 19 passed, 5 skipped
- Chromium data-URI probe — loaded at 300×99

`yarn build:ci` remains blocked by existing `Symbol.dispose` errors in `client/node_modules/@nimiq/core/types/wasm/bundler.d.ts`.

Closes #581

### DIFF
--- a/src/components/LedgerUi.vue
+++ b/src/components/LedgerUi.vue
@@ -525,7 +525,7 @@ export default LedgerUi;
     }
 
     .ledger-screen-confirm-address {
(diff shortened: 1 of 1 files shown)


danimoh (approved):
Thanks.
Note though that this glitch was never actually reachable in the Hub.

PR #583, opened 2026-08-03
Offer a new address from the ChooseAddress picker

Andjroo111:
## The problem

An app that needs a *fresh* address, rather than any existing one, cannot ask for it. It has
to tell the user to leave, open the Nimiq Wallet, use "Add address", name it, come back, and
run `chooseAddress` again. `addAddress` is Nimiq-only by design (`RpcApi.ts:60`), and that is
not what this asks to change.

The case I hit is one address per child in a family allowance app, so the seam repeats per
child on first run. Anything that assigns an address a role rather than reusing the user's
main one has the same shape: an escrow slot, a per-invoice address, a shared household jar.

## What this does

Adds a footer button to `ChooseAddress` that runs the add-address flow already in the Hub and
comes back with the new address selected.

The routing is the pattern the same view already uses for login. `goToOnboarding` sets
`staticStore.originalRouteName = RequestType.CHOOSE_ADDRESS` and pushes `ONBOARD`; `_reply`
then stores the result in `sideResult` and routes back. `addAddress` does the same thing one
route over, so `AddAccount` and `AddAccountSelection` are untouched apart from one fallback.

That precedent is the argument for it. `ChooseAddress` may already send a third-party caller
through `ONBOARD`, which is the one Account Management request explicitly kept off the
whitelist because it exposes internal `accountId`s. Deriving one more address is a smaller
action than creating or importing an account, and `ChooseAddress` resolves with an address and
a label, so no `accountId` is reachable this way.

## Scope

The button appears only when the account to add to is a single unambiguous answer:

- `ui === 2` after an account has been picked, or
- the default UI when exactly one BIP39 account exists

Legacy accounts hold one address, Ledger needs its own flow, and several accounts with no
selection has no correct target. In all of those the button is absent rather than failing
later.

`_3rdPartyRequestWhitelist` and `privilegedOrigins` are not touched. `addAddress` as an RPC
stays exactly as restricted as it is today.

## Two details worth review

**`AddAccount` reads the wallet off the request.** Reached this way it inherits a
`ChooseAddress` request, which carries no `walletId`, so it now falls back to the active
account, which `addAddress` sets before routing. The same change replaces a non-null assertion
on `WalletStore.get` with a rejection, since that lookup can now miss.

**Cancelling in the Keyguard cancels the whole request.** `_reply` deliberately does not honour
`originalRouteName` for `ERROR_CANCELED`, so backing out of the add-address step rejects the
`chooseAddress` call instead of returning to the picker. That matches what the existing
onboarding detour does today, so I left it alone, but returning to the picker is arguably
better for both.

## Verification

`yarn lint` and `yarn build:ci` pass. `yarn i18n:extract` picks up one new string, "Use a new
address", and the regenerated `en.po` is included.

`yarn test` fails here, and it fails the same way on an unmodified checkout: five suites die at
`require('@nimiq/core')` with `ReferenceError: TextDecoder is not defined`, which is the jsdom
environment on my Node, not this change. Same five before and after.

No unit test is added, because the existing suite covers stores and clients rather than views.

**Left as a draft because the flow has not been walked in a browser yet**, which needs a local
Keyguard and a BIP39 account. The parts that are only argued from source so far are the return
path (`sideResult` carrying `Address` back into `created`) and the wallet re-read, which exists
because `AddAccountSelection` persists through `WalletStore` without committing `addWallet`, so
the Vuex copy is stale on the way back. I will mark it ready once it is exercised. Review of
the approach before then is welcome, since that is the part worth arguing about.

### DIFF
--- a/src/i18n/en.po
+++ b/src/i18n/en.po
@@ -716,7 +716,7 @@ msgstr ""
 msgid "Login successful"
 msgstr ""
 
-#: src/views/ChooseAddress.vue:67
+#: src/views/ChooseAddress.vue:73
 #: src/views/ConnectAccount.vue:40
 msgid "Login to another account"
 msgstr ""
@@ -1239,6 +1239,10 @@ msgstr ""
 
 #: src/views/ActivatePolygonSuccess.vue:117
 msgid "USDC/USDT activated"
+msgstr ""
+
+#: src/views/ChooseAddress.vue:40
+msgid "Use a new address"
 msgstr ""
 
 #: src/components/OnboardingMenu.vue:12
--- a/src/views/AddAccount.vue
+++ b/src/views/AddAccount.vue
@@ -4,6 +4,7 @@
 
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator';
+import { State } from 'vuex-class';
 import { DeriveAddressRequest } from '@nimiq/keyguard-client';
 import { BrowserDetection } from '@nimiq/utils';
 import { ParsedSimpleRequest } from '../lib/RequestTypes';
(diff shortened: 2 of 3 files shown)


Andjroo111:
Concrete evidence for the motivation, measured against the deployed testnet Hub today rather
than argued.

An app can already obtain addresses for a purpose without sending the user away, by asking
`connectAccount` for derivation paths. That part works well:

```
asked for m/44'/242'/0'/1' and m/44'/242'/0'/2'
got 2 signatures, both verifying, addresses distinct
no "Add address" step taken first
```

But those addresses are then unspendable through the Hub:

```
signTransaction({ sender: <derived address>, ... })
Error: Address not found
```

which is `SignTransaction.vue:28` resolving the sender through `findWalletByAddress`. So the
addresses exist and belong to the user, and the Hub will not sign from them because they were
never added to the wallet.

That is the gap this PR closes from the other end: adding an address through the picker puts it
in `WalletStore`, so it is both selectable and spendable, without a trip to the wallet app.

PR #586, opened 2026-09-01
Add Github Actions workflow to deploy to testnet

jeffesquivels:


sisou:
@jeffesquivels is this ready for review and testing?

jeffesquivels:
> @jeffesquivels is this ready for review and testing?

Actually, I had a question about this: while testing the deployment workflow, I noticed that the smoke test [would fail](https://github.com/nimiq/hub/actions/runs/33595477183/job/100137943127) unless VUE_APP_HUB_URL [is set](https://github.com/nimiq/hub/pull/586/changes/2cbda4552586cf57b024bee9c33d52b200db023d) when building the hub (which it currently is not). Since this is not set currently, is that a bug on the smoke test or is something else going on?

Thanks!

sisou:
> Since this is not set currently, is that a bug on the smoke test or is something else going on?

Seems it must be set: https://github.com/nimiq/hub/blob/master/vue.config.js#L30

I added that variable when we had a hub.pos... testnet hub deployment for Albatross before the PoS transition, in parallel with the PoW testnet hub. 

That means the env variable should not get defined in the script call, but instead in the[ CI environment](https://github.com/nimiq/hub/settings/environments/21313669284/edit).

PR #588, opened 2026-09-09
Multi-transaction signing for unstaking and validator switches

mraveux:
### Why

Unstaking and switching validators need **several transactions signed in one user interaction**, because
the later ones can only be broadcast after the first has settled on chain — they are then handed to the
watchtower, which sends them when the deactivation matures. `SIGN_TRANSACTION` could only ever carry one
transaction, so the Hub had no way to express the request or to drive a Ledger through several consecutive
signatures.

This is the Hub half. The Keyguard half is T404-16; the wallet half is its own branch.

### What

**Client API** — `client/PublicRequestTypes.ts`, `client/HubApi.ts`

- `SignTransactionRequest` becomes a union of four shapes, with the existing inline single-transaction
  request kept unchanged as `SignTransactionRequestSingle` for backward compatibility:
  - `standard` — a `transactions` array of `TransactionInfo` entries or serialized `Uint8Array`s
  - `switch-validator` — exactly two: set-active-stake, update-staker
  - `unstaking` — exactly three: set-active-stake, retire-stake, remove-stake
- New `TransactionInfo` entry type mirrors the Keyguard's, minus the sender: the **request-level `sender`
  is inherited by every entry**. As in the Keyguard, entries carry no labels; labels stay request-level and
  only apply when there is a single transaction.
- `signTransaction()` now resolves to `SignedTransaction | SignedTransaction[]` — an array exactly when
  more than one transaction was signed.
- `switch-validator` deliberately has **no `validatorAddress`**: the target is derived from the signed
  update-staker's `newDelegation`, so what is displayed is what gets signed. `SignStakingRequest` gains the
  validator address/image and amount fields the wallet needs to render both sides of a switch.

**Request parsing** — new `src/lib/SignTransactionRequestParsing.ts` (703 lines)

- A close port of the Keyguard's request parsing, kept diffable against it check for check, with every
  deviation marked in a comment. This matters because **for Ledger accounts there is no Keyguard**, so
  this parsing plus the Ledger's own display is the only validation the request gets.
- Layout constraints are enforced at parse time (transaction kind and order per layout), along with
  sender/recipient, network id, staking data and label checks.
- Deliberately free of app-graph imports (`RpcApi`, router, views) so it can be imported standalone by
  unit tests — new `tests/unit/SignTransactionRequestParsing.spec.ts`, 65 tests over 1020 lines.
- `rawSignTransactionRequest()` is the inverse mapping for history state, preserving layout and validator
  metadata across a redirect.

**Ledger** — `src/views/SignTransactionLedger.vue` (largely rewritten), `src/components/LedgerUi.vue`

- Arbitrary multi-transaction standard requests, plus the two staking layouts, signed as consecutive Ledger
  interactions.
- A multi-step signing indicator: a checkmark per completed step, the current step number and its
  instructions, pending steps after it — collapsing to a compact `n/total` counter above 8 steps.
- Both validators rendered for a switch, the validator shown on the staking-contract side, zero amounts of
  signalling transactions no longer displayed, and a fix for the network client being unavailable on
  Checkout reload.

**Elsewhere** — `SignTransaction.vue` / `SignStaking.vue` / `SignTransactionSuccess.vue` follow the union
type; `RequestParser` and `RequestTypes` carry the `layout` through; `Cashlink` now uses `cashlink.fee` for
funding as well; multi-transaction signing demos added under `demos/`.

### How it was verified

65 new unit tests on the request parsing. QA then ran the ten-flow test plan on T404-15 for both Keyguard
and Ledger (regular send, validator switch, unstaking, instant payout, start staking, add stake, cashlink
creation, swap refund, standard multi-tx signing, checkout):
https://app.qase.io/public/report/379bd6f17c667627377c8950bcb27604cd5e0505

### Follow-ups

- Released as HubApi v1.15.0 on `master`, after this branch.
- `TransactionInfo.validityStartHeight` stays required until the Hub has its own network connection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ package-lock.json
 *.sw*
 
 deployment-hub
+CLAUDE.md
--- a/client/HubApi.ts
+++ b/client/HubApi.ts
@@ -171,7 +171,7 @@ export default class HubApi<
     public signTransaction<B extends BehaviorType = DB>(
         request: Promise<SignTransactionRequest> | SignTransactionRequest,
         requestBehavior: RequestBehavior<B> = this._defaultBehavior as any,
-    ): Promise<B extends BehaviorType.REDIRECT ? void : SignedTransaction> {
+    ): Promise<B extends BehaviorType.REDIRECT ? void : SignedTransaction | SignedTransaction[]> {
         return this._request(requestBehavior, RequestType.SIGN_TRANSACTION, [request]);
     }
 
--- a/client/PublicRequestTypes.ts
+++ b/client/PublicRequestTypes.ts
@@ -150,6 +150,9 @@ export interface SignTransactionRequestMulti extends BasicRequest {
     sender: string;
     senderLabel?: string;
     recipientLabel?: string;
+    // For `'switch-validator'`, senderLabel and recipientLabel name the validators being
+    // switched between, so the staking address is labelled separately.
+    stakerLabel?: string;
     transactions: TransactionData[] | Uint8Array[];
     validatorAddress?: string;
     validatorImageUrl?: string;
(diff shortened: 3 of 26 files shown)


mraveux:
Already merged into master; opened to keep the change set on record.

